PR #6, opened 2020-11-30
supervisor: Add noise to initial position

mrshu:
* Add some noise to the initial position to which a robot is reset.


mrshu:
cc @Adman

Adman (approved):

PR #14, opened 2020-12-05
refactor referee code into submodule

Adman:
Supervisor - takes care of sending packets etc.
Referee - decides if there are violations of rules, goals etc.


controllers/rcj_soccer_referee_supervisor/referee/referee.py
```diff
@@ -0,0 +1,49 @@
...
+            self.time = self.match_time
+
+        self.draw_time(self.time)
+
+        # If we are currently not in the post-goal waiting period,
+        # check if a goal took place, setup the waiting period and move the
+        # robots to proper positions afterwards.
+        if self.ball_reset_timer == 0:
+            self.check_goal()
+            self.check_robots_in_penalty_area()
```
mrshu, line 41: @Adman I guess we would do this in the `else` part of the statement

mrshu (approved):
Looks good -- we'll have to put together a way of actually dealing with all the checks, especially the "lack of progress/movement" ones.

controllers/rcj_soccer_referee_supervisor/referee/referee.py
Adman: Isn't the `else` part for post-goal waiting period where the goal took place?

mrshu: Nah, you are right @Adman, I didn't get it right away. Please ignore my comment here.

Issue #11, opened 2020-11-30
Lack of progress


Adman:
If the ball does not change its position within some timespan, move the ball to the nearest unoccupied neutral spot.

mrshu:
@Adman to put this together by the specification, #21 needs to be augmented with (at least) the positions of the neutral spots. Plus we ought to have some way of checking whether the neutral spot is occupied or not.

felipenmartins:
I guess this one is solved, right?

mrshu:
@felipenmartins partially -- we still don't have the "find the nearest neutral spot" part figured out but I guess we can close this already and create a new issue to track the progress of this feature.

PR #21, opened 2020-12-14
add: Progress Checker

mrshu:
* Add a ProgressChecker object that ensures the streamlines the process
  of checking progress of various elementes of simulation (i.e. robots
  and ball).

* Refactor position reseting of robots and ball in order for it to be
  effectivelly addressable to a specific robot/ball.

* Check for ball/robot position updates on every tick.

* Handle lack of progress by resetting position.

### DIFF
--- a/controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
+++ b/controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
@@ -1,7 +1,14 @@
+from math import ceil
 from referee.consts import MATCH_TIME, TIME_STEP
 from referee.referee import RCJSoccerReferee
 
-referee = RCJSoccerReferee(match_time=MATCH_TIME)
+referee = RCJSoccerReferee(
+    match_time=MATCH_TIME,
+    progress_check_steps=ceil(15/(TIME_STEP/1000.0)),
+    progress_check_threshold=0.5,
+    ball_progress_check_steps=ceil(10/(TIME_STEP/1000.0)),
+    ball_progress_check_threshold=0.5,
+)
 
 while referee.step(TIME_STEP) != -1:
     referee.emit_positions()
--- a/controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
+++ b/controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
@@ -1,0 +1,44 @@
+import math
+
+
+class ProgressChecker(object):
+    def __init__(self,
+                 steps: int,
+                 threshold: float):
+        self.steps = steps
+        self.threshold = threshold
+        self.prev_position = None
+        self.reset()
+
+    def reset(self):
+        self.A = [0 for _ in range(self.steps)]
+        self.iterator = 0
+
+    def update_position(self, position):
+        if not self.prev_position:
+            self.prev_position = position
+            return
+
+        prev_position = self.prev_position
+
+        delta = math.sqrt((prev_position[0] - position[0]) ** 2 +
+                          (prev_position[2] - position[2]) ** 2)
+
+        self.A[self.iterator % self.steps] = delta
+        self.iterator += 1
+
+        self.prev_position = position
+
+    def is_progress(self, robot=None):
+        s = sum(self.A)
+
+        if robot:
+            print(f'Robot {robot} iterator: {self.iterator}, '
+                  f's: {s}, thr: {self.threshold}')
+        else:
+            print(f'iterator: {self.iterator}, s: {s}, thr: {self.threshold}')
+
+        if self.iterator < self.steps:
+            return True
+
+        return s >= self.threshold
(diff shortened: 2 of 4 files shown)


mrshu:
This should (eventually) resolve #11

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -1,7 +1,14 @@
+from math import ceil
 from referee.consts import MATCH_TIME, TIME_STEP
 from referee.referee import RCJSoccerReferee
 
-referee = RCJSoccerReferee(match_time=MATCH_TIME)
+referee = RCJSoccerReferee(
+    match_time=MATCH_TIME,
+    progress_check_steps=ceil(15/(TIME_STEP/1000.0)),
```
Adman, line 7: What about putting these `ceil(...)` values to consts.py?

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
```diff
@@ -0,0 +1,44 @@
+import math
+
+
+class ProgressChecker(object):
+    def __init__(self,
+                 steps: int,
+                 threshold: float):
```
Adman, line 7:
```python
    def __init__(
        self,
        steps: int,
        threshold: float
    ):
```

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -1,21 +1,23 @@
...
-    N_ROBOTS,
     ROBOT_INITIAL_TRANSLATION,
     ROBOT_INITIAL_ROTATION,
     BALL_INITIAL_TRANSLATION,
 )
 
 
 class RCJSoccerSupervisor(Supervisor):
-    def __init__(self, match_time: int,
+    def __init__(self,
```
Adman, line 15: Same formating here
```python
        def __init__(
            self,
```

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -48,9 +52,17 @@ def __init__(self, match_time: int,
 
             self.robot_in_penalty_counter[robot] = 0
 
+            pc = ProgressChecker(progress_check_steps,
```
Adman, line 55:
```suggestion
            pc = ProgressChecker(
                progress_check_steps,
```

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
```diff
@@ -0,0 +1,44 @@
...
+        self.steps = steps
+        self.threshold = threshold
+        self.prev_position = None
+        self.reset()
+
+    def reset(self):
+        self.A = [0 for _ in range(self.steps)]
```
Adman, line 14: What is `self.A`? We should probably set a self-descriptive name or document it.

Adman (commented):
Overall it looks alright, just a couple of small things to fix

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
mrshu: Cannot say it would look better to me but I guess it is closer to being more "Pythonic" :)

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
mrshu: Ok, this I really do not get. Can you please share a bit of reasoning behind this type of formatting?

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
mrshu: Well, seems to be part of an ongoing discussion: https://github.com/python/typing/issues/433

It's quite tough for me to make up my mind as to what actually looks nicer so let's go with this for the time being.

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
mrshu: Well, an extreme abbreviation of "Array" :)

This is good feedback though, I believe the latest version should make it a bit more explicit.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
mrshu: I must say I don't really like this `consts.py` approach that much. In my view, the functionality of the supervisor ought to be configurable from the `rcj_soccer_referee_supervisor.py` file -- you should not need to update anything else, much less another specific file which is deeper in another repository.

That said, this is certainly a temporary solution as I completely agree that it belongs to a constant, whether in `consts.py` or this particular file.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: this is standard black formating style. But if you don't like it, let's stick with the current formating :)

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
Adman: Ok, no problem

Adman (approved):

PR #23, opened 2020-12-21
Pause simulator at end of match

aquahika:
Since after end of the match simulator is still working and consumpting resources significantly at FAST mode.
Hence pause the simulator after the match and prevent it.


mrshu:
Very nice @aquahika, thanks for putting this together!

Maybe we can just terminate the whole simulation after the time runs out? I really don't know but I guess if pausing the simulation is possible, stopping it altogether should be as well.

aquahika:
Yes, I guess we will, 
But in terms of #24 recording video, we still need a little bit of time to process and export a video before terminating.

mrshu (approved):

mrshu:
Makes a lot of sense -- thanks @aquahika!

I just merged it in, it's a small but very helpful change :)

PR #24, opened 2020-12-21
Add Video Recorder 

aquahika:
Record a game and export as a mp4 file.

(The method @mrshu mentioned before can record a game as HTML 3D format but seems hard to upload to video hosting service)
https://github.com/cyberbotics/webots-animation-action

PS: at FAST mode (run by ctrl-4),  I can run 10 minutes game and export it at **14 seconds** on my environment 😊
Ryzen 7 3700X/ Geforce 1660 Super
It means we can much reduce the processing time! 

https://user-images.githubusercontent.com/2618822/102789327-50fa0100-43e7-11eb-985f-c21f2cced274.mp4

PR #25, opened 2020-12-21
OpenGL-dockerization and JSON-logs through x3d-capture

dbscoach:
I cannot get headless OpenGL to work for the life of me but I do get a full copy of all movements in the game as an x3d-stream by dumping the websocket output to disk

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+#FROM debian:buster
+FROM ubuntu:focal
+EXPOSE 1234
+EXPOSE 42000
+EXPOSE 42001
+EXPOSE 42002
+EXPOSE 42003
+VOLUME /out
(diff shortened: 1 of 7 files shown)


mrshu:
This is a very nice start -- thanks for that @dbscoach!

Just a naive question -- would something similar to https://github.com/cyberbotics/webots-docker be of help here? They claim it is used for GPU accelerated images but who knows what that actually means...

mrshu:
@dbscoach just a friendly ping, given your experiments yesterday, would you happen to have some updates here?

mrshu:
@dbscoach Since @aquahika has merged #24 in, I believe we don't necessarily need to capture all the WebSockets traffic and could live with just being able to run the a simulated game in some sort of a container.

Do you think that would be possible to do relatively quickly?

Thanks!

dbscoach:
Closing in favor of #68 which has both more features and a proper place in a feature branch.

PR #26, opened 2020-12-21
add: JSON logging

### LINKED ISSUE #5: JSON-based game logs
We'll need to have logs on two levels:

1. Debug logs (probably via Python's `logging`)
2. Game logs, which ought to be readable outside of `webots` as well as Python itself

I think one of the better options would be having some JSON lines (https://jsonlines.org/) or something of that sort with action that happened (goal scored, lack of progress, robot's position restarted, ball's position restarted....) as well as some metadata, chiefly among them the timestamp at which the event took place. This would allow us to not just have historical data relevant matches but also to better navigate their recordings (i.e. it should be trivial to see all timestamps at which lack-of-progress took place for instance).

@felipenmartins @javier687 @marcodankel @Adman what do you think

mrshu:
* Add a way of logging events that take place in the referre
  (JSONLogger), which has no dependences outside of Python standard
  library that sends JSON lines to a file.

* Introduce the concept of team names (so that the Supervisor can make
  use of them, in say the logs).

* Put together a way of creating a filename (the `reflog_path` function)

* Fixes #5.


controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -1,13 +1,40 @@
...
+                team_yellow: str):
+    now_str = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+    team_blue = team_blue.replace(' ', '_')
+    team_yellow = team_yellow.replace(' ', '_')
+
+    # Ensure the directory ecists
+    if not directory.exists():
+        directory.mkdir(parents=True, exist_ok=True)
+
+    p = directory / Path(f'{team_blue}_vs_{team_yellow}-{now_str}.json')
```
mrshu, line 23: @aquahika I believe this could be of use for you as well, when it comes to generating the filename of the video.

Let's just do it in the next iteration though, not in the first one :)

controllers/rcj_soccer_referee_supervisor/referee/json_logger.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class JSONLogger(object):
+    def __init__(self,
+                 logfile: Path):
+
+        self.logfile = logfile
+
+    def event(self,
+              supervisor,  # Supervisor from supervisor.py
```
mrshu, line 14: @Adman this is total craziness -- I could not get the type here since there was a circular dependency. I am not sure what's the correct way of dealing with this.

Would you happen to know?

Adman: Try this one
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING
    import Supervisor
...
def event(self, supervisor: 'Supervisor'):
```

controllers/rcj_soccer_referee_supervisor/referee/referee.py
```diff
@@ -25,13 +25,27 @@ def check_progress(self):
             self.progress_chck[robot].track(pos)
 
             if not self.progress_chck[robot].is_progress(robot):
-                print(f'Robot {robot}: Lack of progress')
+                self.log.event(
+                    supervisor=self,
+                    type="lack_of_progress",
```
Adman, line 30: Create special file `choices.py` (or use consts.py if you think it's better). Add Event types there. I would also use uppercase letters for event types.
```python
from enum import Enum

class LogEventChoices(Enum):
     LACK_OF_PROGRESS = "LACK_OF_PROGRESS"
```
```suggestion
                    type=LogEventChoices.LACK_OF_PROGRESS.value,
```

controllers/rcj_soccer_referee_supervisor/referee/json_logger.py
```diff
@@ -0,0 +1,41 @@
+import json
+from pathlib import Path
+from datetime import datetime
+from typing import Optional
+
+
+class JSONLogger(object):
+    def __init__(self,
+                 logfile: Path):
```
Adman, line 9: I kinda don't like this kind of formating. In this case the arguments fit to one line, so I would prefer to have them on the same line. If they do not fit, let's use
```python
def __init__(
    self,  # note only 4 spaces before the argument
    logfile: Path,
):
```

controllers/rcj_soccer_referee_supervisor/referee/json_logger.py
```diff
@@ -0,0 +1,41 @@
...
+            data['robot'] = robot
+            data['team'] = supervisor.robot_to_team_name(robot)
+
+        if team is not None:
+            data['team'] = team
+
+        if payload is not None:
+            data['payload'] = payload
+
+        with self.logfile.open('a') as outfile:
```
Adman, line 39: Does the `append` mode also creates the file in case it does not exist?

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -213,3 +222,11 @@ def reset_robot_position(self, robot):
 
         # Ensure the progress checker does not count this "jump"
         self.progress_chck[robot].reset()
+
+    def robot_to_team_name(self, robot):
```
Adman, line 226:
```suggestion
    def robot_to_team_name(self, robot) -> str:
```

Adman (changes requested):

controllers/rcj_soccer_referee_supervisor/referee/json_logger.py
mrshu: One line it is!

controllers/rcj_soccer_referee_supervisor/referee/json_logger.py
mrshu: @Adman that is indeed my understanding (and what happened in my tests)

mrshu:
Thanks for the review @Adman, I believe I addressed all your comments.

Please feel free to give this another look!

controllers/rcj_soccer_referee_supervisor/referee/referee.py
```diff
@@ -25,13 +26,27 @@ def check_progress(self):
             self.progress_chck[robot].track(pos)
 
             if not self.progress_chck[robot].is_progress(robot):
-                print(f'Robot {robot}: Lack of progress')
+                self.log.event(
+                    supervisor=self,
+                    type=GameEvents.LACK_OF_PROGRESS,
```
Adman, line 31: Hmmm, since `type` is typed as `str`, we should probably call `GameEvents.LACK_OF_PROGRESS.value`

```python
>>> class GameEvents(Enum):
...     A = "AAA"
... 
>>> def foo(x):
...     print(x)
... 
>>> foo(GameEvents.A)
GameEvents.A
>>> foo(GameEvents.A.value)
AAA
```

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -213,3 +222,11 @@ def reset_robot_position(self, robot):
 
         # Ensure the progress checker does not count this "jump"
         self.progress_chck[robot].reset()
+
+    def robot_to_team_name(self, robot) -> str:
```
Adman, line 226: Forgot to type robot.
```suggestion
    def robot_to_team_name(self, robot: str) -> str:
```

controllers/rcj_soccer_referee_supervisor/referee/referee.py
mrshu: Hmm, that's true -- thanks @Adman!

Adman (approved):

PR #27, opened 2020-12-22
A very simple demo game

felipenmartins:
This program implements a very simple game with all robots. It serves as a crude example of how the games will run.

### DIFF
--- a/controllers/rcj_soccer_player/rcj_soccer_player_demo.py
+++ b/controllers/rcj_soccer_player/rcj_soccer_player_demo.py
@@ -1,0 +1,352 @@
+"""rcj_soccer_player controller."""
+from controller import Robot
+import struct
+import math
+
+TIME_STEP = 64
+ROBOT_NAMES = ["B1", "B2", "B3", "Y1", "Y2", "Y3"]
+N_ROBOTS = len(ROBOT_NAMES)
+
+
+def parse_supervisor_msg(packet: str) -> dict:
+    # X, Z and rotation for each robot
+    # plus X and Z for ball
+    struct_fmt = 'ddd' * 6 + 'dd'
+
+    unpacked = struct.unpack(struct_fmt, packet)
+
+    data = {}
+    for i, r in enumerate(ROBOT_NAMES):
+        data[r] = {
+            "x": unpacked[3 * i],
+            "y": unpacked[3 * i + 1],
+            "orientation": unpacked[3 * i + 2]
+        }
+    data["ball"] = {
+        "x": unpacked[3 * N_ROBOTS],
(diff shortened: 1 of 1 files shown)


felipenmartins:
@mrshu and @Adman , could you check if the addition of this example code makes sense? And, if so, make it available in the main branch?
Since we already publicized the github of the soccer sim, I think it would help if teams could see a (simple) game play as a crude example.

Adman:
As I'm now thinking about this, we should probably create 6 separate files, each for one robot. In terms of example program, we could create some Abstract class to put common logic there and instantiate each robot's program with this class. This will require to redefine script names in the world file. So we won't confuse participants that it is put into one file. What do you think about this?

felipenmartins:
> As I'm now thinking about this, we should probably create 6 separate files, each for one robot. In terms of example program, we could create some Abstract class to put common logic there and instantiate each robot's program with this class. This will require to redefine script names in the world file. So we won't confuse participants that it is put into one file. What do you think about this?

Well, you are programmers. So, if you think you can do it, I am all for it! 👍 
I just wanted to give team a simple toy team to play against and use as example.
When do you think you can have this available?

mrshu:
@Adman the way I see it, this would probably be OK, provided that we create that Abstract class--otherwise there is a good chance we may be over-complicating it (and generally raising the bar for entry, which I think would be rather unfortunate).

That said, I do agree that we should aim to create 6 files here, since that is what we'll expect the teams to submit anyway.

@felipenmartins looking at the changes, it seems like most of them amount to changing of variables. I believe we ought to be able to put something similar together on Monday.

felipenmartins:
@mrshu and @Adman Great! Thanks! :-)

Adman:
@mrshu Wouldn't it be better that teams submit entire zip file with code, which would contain 3 files for robots (with names like `robot_a.py`, `robot_b.py` and `robot_c.py` + some helper python scripts/functions located in other python files?

mrshu:
@Adman That would indeed be great!

The only slight update would be that I would like it to be an open Git repository, which we can just `git clone`. The folder structure then could be something on the order of

```bash
$ git clone https://github.com/teamA/simulated-soccer.git teamA
$ tree teamA
teamA/
├── robot1
│   └── robot1.py
├── robot2
│   └── robot2.py
├── robot3
│   └── robot3.py
└── utils
    └── rcj_serializer.py
```

Our automation would then appropriately map this in the Webots world, so that the whole simulation can actually run.

Would that make sense?

mrshu:
Also, it seems like we can close this PR in lieu #30

Adman:
See my comment about directory structure in #30

PR #28, opened 2020-12-23
add inside the penalty area check

### LINKED ISSUE #3: Robots teleported from the penalty area if they are in there for more than 15 seconds
@javier687 @felipenmartins  @Adman Let's see if we can put this together rather quickly

Adman:
* fixes #3


controllers/rcj_soccer_referee_supervisor/referee/referee.py
```diff
@@ -1,25 +1,32 @@
...
+        penalty area for longer period of time
+        """
+        for robot in ROBOT_NAMES:
+            pos = self.robot_translation[robot]
+            self.penalty_area_chck[robot].track(pos, self.time)
+
+            if self.penalty_area_chck[robot].is_violating():
+                print(f'Robot {robot}: Inside penalty area for too long')
+                # TODO: move the robot to the FURTHEST unoccupied neutral spot
+                self.reset_robot_position(robot)
```
Adman, line 22: We will need to find *furthest* unoccupied neutral spot here.

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
```diff
@@ -43,15 +38,14 @@ def track(self, position):
 
         self.prev_position = position
 
-    def is_progress(self, robot=None):
+    def is_progress(self, robot=None) -> bool:
```
mrshu, line 41: @Adman I am wondering whether this ought to actually be a property rather than a function.

What do you think?

In any case, the docstring is missing a `Returns` section, that's something I should probably add on the next pass...

controllers/rcj_soccer_referee_supervisor/referee/referee.py
mrshu: @Adman I guess a prerequisite would be to find out what's the situation like with all the neutral spots (i.e. are their occupied or not). That sounds like a separate module -- the results will probably be useful for other "rules" as well.

What do you think? Worth another PR or will we do it here?

controllers/rcj_soccer_referee_supervisor/referee/progress_checker.py
Adman: It cannot be a property since it takes `robot` argument.

mrshu: @Adman sorry, you are very much right! What (I guess) I was alluding to was a complete overhaul of the architecture, which would allow us to do it the other way around and have things like `robot.is_progressing` (or `self.is_progressing`) and so on.

That will have to wait for the architecture revamp though.

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
```diff
@@ -0,0 +1,73 @@
+from typing import List
+
+from referee.consts import BLUE_PENALTY_AREA, YELLOW_PENALTY_AREA
+
+
+class PenaltyAreaChecker:
+    def __init__(self, allowed: int, reset_after: int, time: int):
```
mrshu, line 7: @Adman would it be better to call this `time_allowed` instead? I had a bit of a hard time parsing out what `allowed` meant.

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
```diff
@@ -0,0 +1,73 @@
...
+        self.allowed = allowed
+        self.reset_after = reset_after
+        self.time = time
```
mrshu, line 10: @Adman let me see if I get this right, this is an attempt at passing a reference of the "current time" from the parent `Supervisor`?

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
```diff
@@ -0,0 +1,73 @@
...
+        self.time_left_penalty = -1
+
+    def is_in_yellow_penalty(self, x: int, z: int) -> bool:
+        return x < self.y_vertical and self.y_lower < z < self.y_upper
+
+    def is_in_blue_penalty(self, x: int, z: int) -> bool:
+        return x > self.b_vertical and self.b_lower < z < self.b_upper
+
+    @property
+    def is_outside_of_penalty_for_longer(self) -> bool:
```
mrshu, line 27: @Adman this is a linguistics nitpick, but should not this be `was_outside_of_penalty_for_longer`? It seems that at the time of calling we are interested in what happened in the past, not the current state.

Feel free to ignore this though -- it's the ultimate nitpick :)

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
```diff
@@ -0,0 +1,73 @@
...
+        return self.time < self.time_left_penalty - self.reset_after
+
+    @property
+    def is_inside_of_penalty_for_long(self) -> bool:
+        return self.time < self.time_entered_penalty - self.allowed
+
+    @property
+    def has_entered(self) -> bool:
+        return self.time_entered_penalty != -1
```
mrshu, line 36: @Adman just a quick thought: do you think extracting `-1` to some constant would make sense here? Probably not, but I still wanted to ask.

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def is_violating(self) -> int:
+        """
+        Detect whether the robot stays for longer period of time inside
+        the penalty area.
+
+        Returns: bool - whether the robot is violating this rule
+        """
+        if self.has_entered and not self.has_left:
+            if self.is_inside_of_penalty_for_long:
```
mrshu, line 69: @Adman looking at the use of this property, I am if something like `is_inside_penalty_over_limit` would not be a better name. Feel free to ignore this though.

mrshu (approved):
Looks good to me, thanks @Adman 

Just a small comment: we will probably need to update the rules in order to reflect the 2 second reset period once the robot leaves the penalty area (cc @marcodankel)

controllers/rcj_soccer_referee_supervisor/referee/referee.py
Adman: @mrshu I would prefer separate PR, where we could add functionality for finding the closest neutral spot.

mrshu: @Adman makes sense, let's get this merged then!

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
Adman: Good catch!

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
Adman: Yes. But it seems useless since the time is remembered in `track` method. Will fix that.

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
Adman: I think we can change `-1` to `None` instead.

controllers/rcj_soccer_referee_supervisor/referee/penalty_area_checker.py
Adman: At the time of calling the robot still `is` outside of penalty area when the method returns `True`. So what about  `has_been_outside_penalty_for_longer` ? :D

Adman:
@mrshu just FYI, I did some code cleanup along my changes.

PR #41, opened 2021-01-02
add: Script for generating specific worlds

mrshu:
* Add a script for generating specific Webots worlds based on some
  configuration. This allows us to change the constollers, images on
  robots, as well as colors of goals without changing the simulation
  logic.

### DIFF
--- a/scripts/generate-soccer-world.py
+++ b/scripts/generate-soccer-world.py
@@ -36,7 +36,7 @@ for option, value in options:
         clean_key = option.replace('--', '')
         params[clean_key] = value
 
-if not template_path.exists():
+if template_path is None or not template_path.exists():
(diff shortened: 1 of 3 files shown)


scripts/generate-soccer-world.py
```diff
@@ -0,0 +1,44 @@
...
+template_path = None
+params = {}
+for option, value in options:
+    if option == '--template':
+        template_path = Path(value)
+    else:
+        clean_key = option.replace('--', '')
+        params[clean_key] = value
+
+if not template_path.exists():
```
Adman, line 39:
```suggestion
if template_path is None or not template_path.exists():
```

Adman (approved):
One small nitpick, otherwise looks good.

mrshu:
Thanks @Adman

PR #42, opened 2021-01-02
Update README.md

felipenmartins:
I included some information to clone the repository also by downloading the ZIP file to make it more friendly to those that don't have Git working on the command line (like me) and not yet familiar with GitHub. 
I also included a link to download Python 3 (just in case). 
Finally, on my computer the command "python" runs Python 3 (probably because I didn't have Python 2 installed before). So, in Webots the correct set to the python command is 'python'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,13 @@ as well as a sample simulated team of robots with some basic strategy.
 
 1. Download [Webots](https://www.cyberbotics.com/#download)
 
-2. Clone this repository to your computer by running
+2. Clone this repository to your computer by downloading the ZIP file from [here](https://github.com/RoboCupJuniorTC/rcj-soccer-sim/archive/master.zip) or running
 
         git clone https://github.com/RoboCupJuniorTC/rcj-soccer-sim.git
 
-3. Make sure the program will be executed by Python version 3. Set `python3` via
-   `Tools > Preferences > Python command`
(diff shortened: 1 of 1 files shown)


mrshu:
Thanks @felipenmartins!

In the future I believe we'll probably need to have a specific set of instructions for each platform (probably just Windows and macOS, possibly even Linux) in the [docs](https://robocupjuniortc.github.io/rcj-soccer-sim/) but this certainly works for now :+1:

README.md
```diff
@@ -14,17 +14,18 @@ as well as a sample simulated team of robots with some basic strategy.
...
+2. Clone this repository to your computer by downloading the ZIP file from [here](https://github.com/RoboCupJuniorTC/rcj-soccer-sim/archive/master.zip) or running
 
         git clone https://github.com/RoboCupJuniorTC/rcj-soccer-sim.git
 
-3. Make sure the program will be executed by Python version 3. Set `python3` via
-   `Tools > Preferences > Python command`
+3. Soccer Sim needs Python version 3.7 (or higher). You can download it from [here](https://www.python.org/downloads/).
 
-4. Use Webots to open the downloaded `soccer.wbt` world located in the `worlds`
+4. In Webots, go to `Tools > Preferences > Python command` and set it to `python` or `python3`.
```
Adman, line 23: You mention that `python` command is actually python3 on your computer. So I would try to explain that `python` can sometimes reference python version 2 and users need to be careful which one is it referencing to.

PR #43, opened 2021-01-03
Added info about Python 3 in Readme

felipenmartins:
Added info about Webots configuration to work with Python 3, according to @Adman ‘s suggestion.


Adman (approved):
Thanks a lot! :)

PR #44, opened 2021-01-03
update: add event handler for displaying messages

Adman:
* introduce event reader which fires event to
  subscribed event handlers


Adman:
This is what the messages look like:

image

controllers/rcj_soccer_referee_supervisor/referee/consts.py
```diff
@@ -62,6 +62,26 @@ class Team(Enum):
...
+    EVENT_MESSAGE_4 = 13
+    EVENT_MESSAGE_5 = 14
+    EVENT_MESSAGE_6 = 15
+    EVENT_MESSAGE_7 = 16
+    EVENT_MESSAGE_8 = 17
+    EVENT_MESSAGE_9 = 18
+    EVENT_MESSAGE_10 = 19
+
 
 class GameEvents(Enum):
```
Adman, line 86: @mrshu I think we can get rid of this Enum, since we are not using its features and make it similar to LabelIds class

mrshu: @Adman I guess `Enum` gives us things like `GameEvents['MATCH_START']` for free, so I don't know. You are probably right though, there is a very good chance we won't use any of these features.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -111,60 +118,99 @@ def _update_positions(self):
...
+            message = "{} - {}".format(time_to_string(time), msg)
+            self.setLabel(
+                getattr(LabelIds, "EVENT_MESSAGE_{}".format(i)),
+                message,
+                0.01,
+                0.95 - ((i-1)*0.02),
+                0.05,
+                0xffffff,
+                0.0,
+                "Arial"
```
mrshu, line 191: @Adman let's try `Lucida Console` in here, just for the fun of it.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -111,60 +118,99 @@ def _update_positions(self):
...
+            "Arial",
+        )
+
+    def draw_event_messages(self):
+        """Visualize (draw) the event messages from queue"""
+        rev = reversed(self.event_messages_to_draw)
+        for i, (time, msg) in enumerate(rev, start=1):
+            message = "{} - {}".format(time_to_string(time), msg)
+            self.setLabel(
+                getattr(LabelIds, "EVENT_MESSAGE_{}".format(i)),
```
mrshu, line 184: @Adman is this actually better than having an `Enum` and doing something like

```python
LableIds[f"EVENT_MESSAGE_{i}"]
```

I honestly don't know, but I guess you'll know :)

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -111,60 +118,99 @@ def _update_positions(self):
...
 
     def _pack_packet(
         self,
         robot_rotation: dict,
         robot_translation: dict,
         ball_translation: list,
     ):
-        """
-        Take the positions and rotations of the robots and the ball and pack
+        """ Take the positions and rotations of the robots and the ball and pack
```
mrshu, line 206: @Adman I realize this is Google's own suggestion for their docstrings but I cannot help it, it looks way worse than at its own line :)

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -111,60 +118,99 @@ def _update_positions(self):
...
+            0.1,
+            0x000000,
+            0.0,
+            "Arial",
+        )
+
+    def draw_event_messages(self):
+        """Visualize (draw) the event messages from queue"""
+        rev = reversed(self.event_messages_to_draw)
+        for i, (time, msg) in enumerate(rev, start=1):
```
mrshu, line 181: @Adman just out of curiosity, do you see any advantages of doing this vs. having a single, `\n` separated string? I guess this way allows the messages to go from the bottom up, no matter how many of them are in the queue/list, right?

controllers/rcj_soccer_referee_supervisor/referee/referee.py
```diff
@@ -125,44 +121,46 @@ def kickoff(self, team: Optional[str] = None):
...
 
-        self.log.event(
+        self.eventer.event(
             supervisor=self,
             type=GameEvents.KICKOFF.value,
-            robot_name=robot_name,
-            msg=f"Robot {robot_name} is kicking off."
+            payload={
+                "robot_name": robot_name,
+                "team": team,
```
mrshu, line 129: @Adman this is my bad, but to be consistent, let's call this `team_name` as well (just as above)

mrshu (approved):
Nicely done @Adman!

I only left some very small suggestions here and there.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
```diff
@@ -0,0 +1,109 @@
...
+        return f"Robot {robot_name} is kicking off."
+
+    def create_match_start_msg(self, total_match_time: int, **kwargs) -> str:
+        return f"The match ({total_match_time}s) has started."
+
+    def create_match_finish_msg(self, total_match_time: int, **kwargs) -> str:
+        return f"The match time {total_match_time}s is over."
+
+    def handle(self, supervisor, type: str, payload: Optional[dict] = None):
+        event_mapping = {
```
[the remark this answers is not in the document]
Adman: We somehow need to map event to method which creates the resulting message.
The other way could be 

```python
def create_event_type_msg():
    ...

method = getattr(self, f"create_{EVENT_TYPE.lower()}_msg")
```

Would that be

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
```diff
@@ -0,0 +1,109 @@
...
+            GameEvents.INSIDE_PENALTY_FOR_TOO_LONG.value: {
```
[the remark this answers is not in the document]
Adman: I'm not sure what you mean by "get rid of these values". Do you want to remove `.value`?

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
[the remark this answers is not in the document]
mrshu: @Adman Right, I see. If all we do is this mapping (i.e. there are no additional arguments to be passed), I guess it may be a bit more straightforward.

I don't have a very well formed opinion here though.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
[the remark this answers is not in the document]
mrshu: @Adman sorry, yes, I meant that `.value` is rather redundant here (the value is basically irrelevant, as it never gets used in its raw form).

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: We limit the number of messages displayed by `MAX_EVENT_MESSAGES_IN_QUEUE`. But I think using `\n` is better in a way that we don't have to have multiple `EVENT_MESSAGE_id` IDs, only one. Will rework that

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Will use Tahoma. Lucid Console is way too wide.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Now that we are using `\n` we need only one label.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
```diff
@@ -0,0 +1,109 @@
...
+    def handle(
+        self,
+        supervisor,  # Supervisor from supervisor.py
+        type: str,
+        payload: Optional[dict] = None,
+    ):
+        """Handle the incoming event
+
+        Args:
+            supervisor (Supervisor): Referee supervisor
```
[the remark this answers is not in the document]
Adman: Fixed

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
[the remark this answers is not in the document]
Adman: Done.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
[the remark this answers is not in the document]
Adman: Completely removed.

controllers/rcj_soccer_referee_supervisor/referee/consts.py
```diff
@@ -62,6 +62,26 @@ class Team(Enum):
 YELLOW_PENALTY_AREA = (-0.59, -0.35, 0.35)
 BLUE_PENALTY_AREA = (0.59, -0.35, 0.35)
 
+MAX_EVENT_MESSAGES_IN_QUEUE = 10
+
+
+class LabelIds:
```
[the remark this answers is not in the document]
Adman: Added docstring.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Done.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Refactored, use `\n` and only one label.

controllers/rcj_soccer_referee_supervisor/referee/referee.py
Adman: Renamed.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
```diff
@@ -0,0 +1,89 @@
...
+        return f"Robot {robot_name} is kicking off."
+
+    def create_match_start_msg(self, total_match_time: int, **kwargs) -> str:
+        return f"The match ({total_match_time}s) has started."
+
+    def create_match_finish_msg(self, total_match_time: int, **kwargs) -> str:
+        return f"The match time {total_match_time}s is over."
+
+    def handle(self, supervisor, type: str, payload: Optional[dict] = None):
+        msg_creator = getattr(self, f"create_{type.lower()}_msg")
```
mrshu, line 86: @Adman this ends up being very nice!

Can we get a bit of a comment here about how it works? Just for the future's sake :)

mrshu: Furthermore, I think a better name would be `msg_formatter`

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -111,60 +118,101 @@ def _update_positions(self):
...
+            "Arial",
+        )
+
+    def draw_event_messages(self):
+        """Visualize (draw) the event messages from queue"""
+        messages = []
+        for time, msg in self.event_messages_to_draw:
+            messages.append("{} - {}".format(time_to_string(time), msg))
+
+        if messages:
```
mrshu, line 184: @Adman it seems like currently the messages are listed here basically forever but I guess we'd want something on the order of "message disappears after X seconds", where X is configurable. That can certainly wait for another PR though :)

mrshu (approved):
Thanks @Adman, just some very, very small nitpicks. Looks good to go to me!

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Yes, let's check and discuss tonight what would look nicer and fix it in another PR if needed.

controllers/rcj_soccer_referee_supervisor/referee/event_handlers.py
Adman: Done.

PR #53, opened 2021-01-11
add: Team names to the screen

### LINKED ISSUE #47: Add labels with team names to the video screen
So that it's not just blue/yellow numbers in the corners

mrshu:
* Ensure the team names are being displayed on the screen.

* Fixes #47.

### DIFF
--- a/controllers/rcj_soccer_referee_supervisor/referee/consts.py
+++ b/controllers/rcj_soccer_referee_supervisor/referee/consts.py
@@ -71,6 +71,8 @@ class LabelIDs(Enum):
     YELLOW_SCORE = 1
     TIME = 2
     EVENT_MESSAGES = 3
+    BLUE_TEAM = 10
+    YELLOW_TEAM = 11
 
 
 class GameEvents(Enum):
--- a/controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
+++ b/controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
@@ -147,6 +147,17 @@ class RCJSoccerSupervisor(Supervisor):
         )
 
         self.setLabel(
+            LabelIDs.BLUE_TEAM.value,
+            self.team_name_blue,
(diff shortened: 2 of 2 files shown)


controllers/rcj_soccer_referee_supervisor/referee/consts.py
```diff
@@ -71,6 +71,8 @@ class LabelIDs(Enum):
     YELLOW_SCORE = 1
     TIME = 2
     EVENT_MESSAGES = 3
+    BLUE_TEAM = 10
+    YELLOW_TEAM = 11
```
Adman, line 75: Why 10 and 11?

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -146,6 +146,17 @@ def draw_scores(self, blue: int, yellow: int):
             "Tahoma",  # Font
         )
 
+        self.setLabel(
```
Adman, line 149: Do we need to be rewriting the team names in each tick? Cannot it happen once?

Adman (approved):
I have just some questions, otherwise LGTM

controllers/rcj_soccer_referee_supervisor/referee/consts.py
mrshu: Just to be on the safe side (since the `GOAL` was 4 in the other PR).

But we can easily refactor this.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
mrshu: It probably does not need to but this felt safer. But I can try to extract it out if you find that to be a show stopper.

mrshu:
@Adman Would you mind taking one more look? I think you'll like this a bit more :)

Adman (approved):

PR #56, opened 2021-01-14
fix: replacing deprecated functions

fcamichel:
Since Webots version 2021a the following functions are deprecated:
- getMotor
- getReceiver
- getEmitter

They can all be replaced with the function getDevice.

webots_deprecated_warnings


mrshu:
Thanks for fixing it @fcamichel, I meant to do so one day but you've beat me to it :)

PR #59, opened 2021-01-20
More detailed installation instructions

felipenmartins:
Added more details to (hopefully) facilitate the installation by beginners.


Adman:
@mrshu I think we should link the README.md to docs/docs/index.md so we do not have to update the same thing twice.

Adman (approved):

mrshu:
This does need to get moved to the docs indeed. I'll try to do that in a subsequent PR.

Thanks for putting this together @felipenmartins !

mrshu (approved):

PR #61, opened 2021-01-21
Change 6 to N_ROBOTS.

TBSDrJ:
In trying to get a simulator working with only one robot, I found two places where 6 was used instead of N_ROBOTS.


Adman:
Thanks for the change!

PR #62, opened 2021-01-22
docs: add hello world robot documentation

Adman:

### DIFF
--- a/docs/docs/how_to.md
+++ b/docs/docs/how_to.md
@@ -34,6 +34,8 @@ N_ROBOTS = len(ROBOT_NAMES)
 class MyRobot:
     def __init__(self):
         self.robot = Robot()
+        self.name = self.robot.getName()
+
         self.receiver = self.robot.getDevice("receiver")
         self.receiver.enable(TIME_STEP)
 
@@ -121,6 +123,7 @@ going to continue with our OOP approach.
 ```python
 def __init__(self):
     self.robot = Robot()
+    self.name = self.robot.getName()
     ...
 ```
 
@@ -128,6 +131,9 @@ The `__init__` something like constructor of the class and is called when the ob
 is created. We initialize some important variables there. The most important one
 is the `Robot` instance, which allows us to get access to the so called devices
 like motor (for controlling the speed) or receiver (for reading data from supervisor).
+**The name and the team of your robot** can be determined by calling `self.robot.getName()`.
+It will give you one of `{"B1", "B2", "B3", "Y1", "Y2", "Y3"}`. The first letter determines
+the team ("Blue", "Yellow"), while the second one is the robot's identifier.
 
 ```python
 def get_new_data(self):
@@ -189,7 +195,6 @@ and its actually a good practice not to duplicate code all over the place.
 Imagine you have the folder structure which looks like this
 
 ```
-controllers
 controllers/
 ├── robot1/
 │   └── robot1.py
--- a/docs/docs/how_to.md
+++ b/docs/docs/how_to.md
@@ -8,8 +8,8 @@ one controller, implemented as a **Python3** program. There is an invisible refe
 which takes care of controlling the game and checking the rules.
 
 The controllers are located in the `controllers` directory. The name of the controller
-must be located in a subfolder with the same name and this name ought to be specified
-in `soccer.wbt` file.
+must be located in a subfolder with the same name (i.e. `my_controller/my_controller.py`)
+and this name ought to be specified in `soccer.wbt` file.
 
 ## Hello world, robot!
 
@@ -225,3 +225,5 @@ magic lines to `robot1.py`, but instead just call
 ```python
 import utils
 ```
+
+because the PATH actually contains the folder of the script which is being run.
--- a/docs/docs/how_to.md
+++ b/docs/docs/how_to.md
@@ -1,0 +1,222 @@
+# How to program your robot
+
+## Controllers
+
+Each object in the simulation world is able to be controlled by a program.
+This program is called **Controller**. Each of the robots should have exactly
+one controller **Python3** program. There is an invisible referee object in the simulation,
+which takes care of controlling the game and checking the rules.
+
+The controllers are located in `controllers` directory. The name of the controller
+must be located in a subfolder with the same name and this name ought to be specified
+in `soccer.wbt` file.
+
+## Hello world, robot!
+
+We have prepared an example robot controllers, which can be found in `controllers`
+directory prefixed with `rcj_soccer_player`. Each robot is programmed the same way,
+meaning it rotates to face the ball and goes towards it afterwards.
+
+We decided to make the code reusable and put some common methods into
+`RCJSoccerRobot` class inside `rcj_soccer_robot.py` file.
+
+Let's put together a simple program on how to program the robot.
+
+```python
+from controller import Robot
+import struct
+
+TIME_STEP = 64
+ROBOT_NAMES = ["B1", "B2", "B3", "Y1", "Y2", "Y3"]
+N_ROBOTS = len(ROBOT_NAMES)
+
+
+class MyRobot:
+    def __init__(self):
+        self.robot = Robot()
+        self.receiver = self.robot.getDevice("receiver")
+        self.receiver.enable(TIME_STEP)
+
+        self.left_motor = self.robot.getDevice("left wheel motor")
+        self.right_motor = self.robot.getDevice("right wheel motor")
+
+        self.left_motor.setPosition(float('+inf'))
+        self.right_motor.setPosition(float('+inf'))
+
+        self.left_motor.setVelocity(0.0)
+        self.right_motor.setVelocity(0.0)
+    
+    def get_new_data(self):
+        packet = self.receiver.getData()
+        self.receiver.nextPacket()
+
+        struct_fmt = 'ddd' * 6 + 'dd'
+
+        unpacked = struct.unpack(struct_fmt, packet)
+
+        data = {}
+        for i, r in enumerate(ROBOT_NAMES):
+            data[r] = {
+                "x": unpacked[3 * i],
+                "y": unpacked[3 * i + 1],
+                "orientation": unpacked[3 * i + 2]
+            }
+        data["ball"] = {
+            "x": unpacked[3 * N_ROBOTS],
+            "y": unpacked[3 * N_ROBOTS + 1]
+        }
+        return data
+
+    def run(self):
+        while self.robot.step(TIME_STEP) != -1:
+            if self.receiver.getQueueLength() > 0:
+                data = self.get_new_data()
+
+                # Get the position of our robot
+                robot_pos = data[self.name]
+                # Get the position of the ball
+                ball_pos = data['ball']
+
+                self.left_motor.setVelocity(1)
+                self.right_motor.setVelocity(-1)
+
+
+my_robot = MyRobot()
+my_robot.run()
+```
+
+Let's explain the code in detail:
+
+```python
+from controller import Robot
+```
+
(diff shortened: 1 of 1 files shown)


felipenmartins (approved):
Nice! I think we should publish it. I already got a question from a Brazilian team (in Portuguese) that would be answered by this.

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+## Hello world, robot!
+
+We have prepared an example robot controllers, which can be found in `controllers`
+directory prefixed with `rcj_soccer_player`. Each robot is programmed the same way,
+meaning it rotates to face the ball and goes towards it afterwards.
+
+We decided to make the code reusable and put some common methods into
+`RCJSoccerRobot` class inside `rcj_soccer_robot.py` file.
+
+Let's put together a simple program on how to program the robot.
```
mrshu, line 23:
```suggestion
Let's put together a simple program to showcase how you can go about programming a robot.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+```python
```
mrshu, line 25: @Adman this is quite a big jump, I though we'd go line by line or something like that. Is that missing by any chance?

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+                self.left_motor.setVelocity(1)
+                self.right_motor.setVelocity(-1)
+
+
+my_robot = MyRobot()
+my_robot.run()
+```
+
+Let's explain the code in detail:
```
mrshu, line 90: @Adman I believe it may be better to start with this and then put the whole code block towards the end.

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+```python
+from controller import Robot
+```
+
+The Robot class is required to be imported because this is the only way we are able
+to controll the robot. The Robot class is shipped together with webots.
```
mrshu, line 97:
```suggestion
to controll the robot. The Robot class is shipped together with Webots.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+```python
+import struct
+```
+
+This library is a built-in python library, which is required to unpack the data sent by
```
mrshu, line 103:
```suggestion
This library is a [built-in Python library](https://docs.python.org/3/library/struct.html), which is required to unpack the data sent by
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+Let's explain the code in detail:
+
+```python
+from controller import Robot
+```
+
+The Robot class is required to be imported because this is the only way we are able
+to controll the robot. The Robot class is shipped together with webots.
+
+```python
```
mrshu, line 99: @Adman do we actually need to import it in here? Wouldn't it be sufficient to import it where the unpacking actually happens?

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+This library is a built-in python library, which is required to unpack the data sent by
+the supervisor.
+
+```python
+TIME_STEP = 64
+ROBOT_NAMES = ["B1", "B2", "B3", "Y1", "Y2", "Y3"]
+N_ROBOTS = len(ROBOT_NAMES)
+```
+
+We define some constants here, whose values we will be used later.
```
mrshu, line 112:
```suggestion
We also define some useful constants, whose values will be used later.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+```python
+class MyRobot:
+```
+
+You can wrap the program into the class as we did. The benefit of OOP
+(object oriented programmin) is that you can later inherit from common class used
```
mrshu, line 119:
```suggestion
([Object Oriented Programming](https://realpython.com/python3-object-oriented-programming/)) is that you can later reuse the same common class
```

mrshu: @Adman I don't really like the cheap advertisement but it seems like a pretty nice link in this case

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+going to continue with our OOP approach.
+
+```python
+def __init__(self):
+    self.robot = Robot()
+    self.name = self.robot.getName()
+    ...
+```
+
+The `__init__` something like constructor of the class and is called when the object
```
mrshu, line 130:
```suggestion
The `__init__` function is something like a constructor of the class and is called when an instance of the `MyRobot` object
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+is created. We initialize some important variables there. The most important one
```
mrshu, line 131:
```suggestion
is created. We use this function to initialize some important variables. The most important one
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+is the `Robot` instance, which allows us to get access to the so called devices
```
mrshu, line 132:
```suggestion
is the `Robot` instance, which allows us to get access to the so called Webots devices
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+like motor (for controlling the speed) or receiver (for reading data from supervisor).
```
mrshu, line 133:
```suggestion
like motor (for controlling the speed) or receiver (for reading data from Supervisor).
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+**The `step` function is crucial and must be used in every controller.
+This function synchronizes the sensor and actuator data between Webots and the controllers.**
+
+
+```python
+if self.receiver.getQueueLength() > 0:
+```
+
+Before reading data, its important to check if there is actually something
```
mrshu, line 165:
```suggestion
Before reading data, it is important to check if there is actually something
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+in the queue.
```
mrshu, line 166:
```suggestion
in the queue that we can read.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+data = self.get_new_data()
+
+robot_pos = data[self.name]
+ball_pos = data['ball']
+
+self.left_motor.setVelocity(1)
+self.right_motor.setVelocity(-1)
+```
+
+And finally, read the new data received from supervisor, do some calculations
```
mrshu, line 178:
```suggestion
And finally, after reading the new data received from supervisor, we do some calculations
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+and set the speed of the motors.
+
+```python
+my_robot = MyRobot()
+my_robot.run()
+```
+
+In the very end of our program, we initialize the robot and call the method `run`
+in order to run the code.
```
mrshu, line 187:
```suggestion
in order to execute the code.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+## Importing shared code
+
+Each team consists out of three robots. These robots might share some of the code
```
mrshu, line 192:
```suggestion
Each team consists of three robots. These robots might share some of the code
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+and its actually a good practice not to duplicate code all over the place.
```
mrshu, line 193:
```suggestion
and it is  actually [a good practice](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) not to duplicate code all over the place.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+├── robot1/
+│   └── robot1.py
+|   └── utils.py
+├── robot2/
+│   └── robot2.py
+├── robot3/
+│   └── robot3.py
+```
+
+and within `robot2.py` you want to import some useful code from the `utils.py`
```
mrshu, line 208:
```suggestion
and within `robot2.py` you want to import some useful code from `utils.py`
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+located in `robot1` folder. This is possible, but you need to put the code below
```
mrshu, line 209:
```suggestion
located in the `robot1` folder. This is possible, but you need to put the follow code snippet right at the
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+on top of your `robot2.py` file.
```
mrshu, line 210:
```suggestion
top of your `robot2.py` file.
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+```python
+import sys
+from pathlib import Path
+sys.path.append(str(Path('.').absolute().parent))
+from robot1 import utils
+``` 
+
+This will ensure that the code from `robot1` is importable (adds the parent folder
```
mrshu, line 219:
```suggestion
This will ensure that the code from `robot1` is importable (it adds the parent folder
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+to the PATH)
```
mrshu, line 220:
```suggestion
to the Python PATH -- the list of folders where Python looks when importing modules).
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+If you want to import `utils` within the `robot1`, you do no have to add these
```
mrshu, line 222:
```suggestion
If you want to import `utils` within `robot1`, you do no have to add these
```

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+magic lines to `robot1.py`, but instead just call
+
+```python
+import utils
+```
```
mrshu, line 227: @Adman it would be great to add a bit of a background so as to why is that the case

mrshu (approved):
I left a ton of nitpicks here and there. If you think it may be worth it, feel free to address them but the docs are already quite readable as is, so we can just get it merged as well :)

Thanks again for doing this @Adman !

docs/docs/how_to.md
Adman:
```suggestion
located in the `robot1` folder. This is possible, but you need to put the following code snippet right at the 
```

docs/docs/how_to.md
Adman: Yeah, line by line is explained below.

docs/docs/how_to.md
Adman: It would, but I think that's not a good practice to import stuff in functions (unless you have cyclic import problem)

docs/docs/how_to.md
Adman: Added.

docs/docs/how_to.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Controllers
+
+Each object in the simulation world is able to be controlled by a program.
+This program is called **Controller**. Each of the robots should have exactly
+one controller **Python3** program. There is an invisible referee object in the simulation,
+which takes care of controlling the game and checking the rules.
+
+The controllers are located in `controllers` directory. The name of the controller
+must be located in a subfolder with the same name and this name ought to be specified
```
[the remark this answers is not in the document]
Adman: Included.

Issue #55, opened 2021-01-13
Wheels slowly drifting away from the robot


Bluskript:
![Peek_2020-12-27_15-29](https://user-images.githubusercontent.com/52386117/104470245-82e62980-5587-11eb-9e3d-9932d767578b.gif)

I don't know what has been causing this, it might be a bug related to rotation since I tweaked those values. I'll report any findings related to this bug if I find any.

Bluskript:
found minimal reproducible test case:
```diff
# rcj_soccer_player controller - ROBOT Y3

###### REQUIRED in order to import files from Y1 controller
import sys
from pathlib import Path

sys.path.append(str(Path(".").absolute().parent))
# You can now import scripts that you put into the folder with your
# robot B1 controller
from rcj_soccer_player_y1 import rcj_soccer_robot, utils

######

# Feel free to import built-in libraries
import math


class MyRobot(rcj_soccer_robot.RCJSoccerRobot):
    def run(self):
        while self.robot.step(rcj_soccer_robot.TIME_STEP) != -1:
            if self.is_new_data():
                data = self.get_new_data()

                # Get the position of our robot
                robot_pos = data[self.name]
                # Get the position of the ball
                ball_pos = data["ball"]

                # Get angle between the robot and the ball
                # and between the robot and the north
                ball_angle, robot_angle = self.get_angles(ball_pos, robot_pos)

                # Compute the speed for motors
-                 direction = utils.get_direction(ball_angle)
+                direction = 1

                # If the robot has the ball right in front of it, go forward,
                # rotate otherwise
                if direction == 0:
                    left_speed = -5
                    right_speed = -5
                else:
                    left_speed = direction * 4
                    right_speed = direction * -4

                # Set the speed to motors
                self.left_motor.setVelocity(left_speed)
                self.right_motor.setVelocity(right_speed)


my_robot = MyRobot()
my_robot.run()
```

Adman:
Hello @Bluskript. Could you please test if this problem occurs with the newest software?

I'll close this issue for now. If you ran into similar problem like you reported, feel free to reopen this issue.

PR #64, opened 2021-01-25
update: find nearest/furthest unoccupied neutral spot

### LINKED ISSUE #45: "Find furthest/nearest unoccupied neutral spot"
In order to implement rules

- 1.8 Inside the Penalty Area
- 1.9 Lack of Progress
- 1.10 Damaged Robots

We need a way of finding the nearest/furthest unoccupied neutral spot. It would be best to abstract this functionality out so that it can be used from various submodules (such as the penalty area/lack of progress checker(s)).

cc @Adman @felipenmartins

Adman:
* closes #45


controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -341,27 +388,31 @@ def _add_initial_position_noise(
...
+            object_name (str): Either "bal" or the robot's name.
+        """
+        self.progress_chck[object_name].reset()
+        if object_name != 'ball':
+            self.penalty_area_chck[object_name].reset()
 
     def robot_name_to_team_name(self, robot_name: str) -> str:
         if robot_name.startswith('Y'):
-            return self.team_name_yellow
+            return Team.YELLOW.value
```
mrshu, line 413: @Adman I don't think this is right. The purpose of this function is to return the actual name of the team so that it could get logged into the reflog and so on -- I suppose `Team.BLUE.value` will end up being the same in all runs.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -316,18 +322,59 @@ def reset_positions(self):
...
-        ball_translation_field.setSFVec3f(BALL_INITIAL_TRANSLATION)
+        ball_translation_field.setSFVec3f(position)
+        self.reset_ball_velocity()
+        self.ball_translation = position
 
+    def reset_ball_velocity(self):
+        """Reset the ball's velocity."""
         self.ball.setVelocity([0, 0, 0, 0, 0, 0])
+
+    def reset_ball_position(self):
```
mrshu, line 375: @Adman I think we should also call `resetPhysics` in here (as well as in the `reset_robot_position`) -- here is the webots issue which discusses what our issue #55  also noticed

https://github.com/cyberbotics/webots/issues/1118

mrshu (approved):
A few small nitpicks but otherwise this looks good to me!

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Yeah, sorry, will revert it back.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Thanks, should be fixed. @mrshu could you please recheck again?

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -340,22 +407,15 @@ def _add_initial_position_noise(
...
 
-        rot_field = self.getFromDef(robot).getField('rotation')
-        rot_field.setSFRotation(ROBOT_INITIAL_ROTATION[robot])
-
-        # Ensure the progress checker does not count this "jump"
-        self.progress_chck[robot].reset()
-
-        self.penalty_area_chck[robot].reset()
+        Args:
+            object_name (str): Either "bal" or the robot's name.
```
mrshu, line 414:
```suggestion
            object_name (str): Either "ball" or the robot's name.
```

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -376,10 +436,95 @@ def reset_team_for_kickoff(self, team: str):
...
+            if is_better and not self.is_neutral_spot_occupied(ns_x, ns_z):
+                best_distance = spot_distance
+                best_spot = ns
+
+        return best_spot
+
+    def move_object_to_neutral_spot(
+        self,
+        object_name: str,
+        neutral_spot: Optional[str],
```
mrshu, line 513:
```suggestion
        neutral_spot: Optional[str] = None,
```

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -376,10 +436,95 @@ def reset_team_for_kickoff(self, team: str):
...
+    ):
+        """Move the robot to the specified neutral spot.
+
+        Args:
+            object_name (str): Name of the object (Ball or robot's name)
+            neutral_spot (str, optional): The spot the robot will be moved to
+        """
+        if neutral_spot is None:
```
mrshu, line 521: @Adman I don't see this function being used without the neutral spot provided, nor does it seem like it would actually do anything, if `neutral_spot` was not provided.

Do you have some future plans with this function?

mrshu (approved):
Thanks @Adman, I found some small nitpicks but other than that, this looks good to me.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: I think it is ok not to have default value here.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: @mrshu There is one edge case. Imagine all the robots and ball are located on neutral spots, so you cannot move the object anywhere.

PR #68, opened 2021-02-01
Barebones Team-vs-Team-Proof of Concept in Docker containers

dbscoach:
It's fairly basic but if you have Open Source GPU Drivers or nvidia-docker set up correctly you can run 

    TEAM_Y=001 TEAM_B=002 ./run-docker.sh

and webots will run (on your host's Xorg-session but without internet access for the container) one 10-minute-Halftime while returning the referee-log-json and mp4-recording of the game to ~/webots_video_out/


Adman (changes requested):
Nice! Some small changes required, mostly to make the code more readable :)

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -28,6 +29,8 @@ def output_path(
         directory.mkdir(parents=True, exist_ok=True)
 
     p = directory / Path(f'{team_blue}_vs_{team_yellow}-{now_str}')
```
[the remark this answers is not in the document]
dbscoach: I didn't touch that one for lack of understanding, I just stuck my output in where it looked plausible to override... Do you mean we should move the entire function down the file? Maybe we should consider doing that later and/or on master to avoid weird merge conflicts.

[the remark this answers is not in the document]
Adman: I just wanted to make the function more readable.

```python
def output_path( 
    directory: Path,
    team_blue: str,
    team_yellow: str,
) -> PosixPath:
    now_str = datetime.utcnow().strftime('%Y%m%dT%H%M%S.%fZ')
    team_blue = team_blue.replace(' ', '_')
    team_yellow = team_yellow.replace(' ', '_')

    if automatic_mode:
        directory = Path('/out/')

    # Ensure the directory ecists
    if not directory.exists():
         directory.mkdir(parents=True, exist_ok=True)

    return directory / Path(f'{team_blue}_vs_{team_yellow}-{now_str}')
```

run-docker.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+./x11docker/x11docker --gpu --runtime=nvidia --network=none --verbose -- \
+-v ${HOME}/webots_video_out/:/out \
+-v $(pwd)/teams/${TEAM_Y}/bot1/bot1.py:/app/controllers/rcj_soccer_player_y1/rcj_soccer_player_y1.py:ro \
```
mrshu, line 4: @dbscoach I'd very much prefer us map the folders rather the specific Python scripts, just for the simplicity of it :)

run-webots.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+webots --mode=fast --batch /app/worlds/soccer.wbt
```
mrshu, line 2: Just out of curiosity, what does `--batch` do?

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:focal
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install wget gdebi-core xvfb supervisor ffmpeg x264 aria2 xserver-xorg-video-all mesa-utils tigervnc-standalone-server python3 python3-pip xdotool xrdp x11-apps x11-common x11-session-utils x11-xkb-utils x11-xserver-utils x11proto-core-dev x11proto-dev xserver-common xserver-xorg-core xserver-xorg-video-nvidia-455 libxv1 x11-xserver-utils && pip3 install websockets
+RUN aria2c -x 16 "https://github.com/cyberbotics/webots/releases/download/R2021a/webots_2021a_amd64.deb" && gdebi -n webots_2021a_amd64.deb && rm webots_2021a_amd64.deb
+EXPOSE 1234
```
mrshu, line 4: @dbscoach Just out of curiosity, is there any particular reason behind exposing these ports in particular?

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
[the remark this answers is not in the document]
mrshu: Arguably, this should probably not happen inside the function at all -- we should make it so that the `directory` parameters that is passed to `output_path` is already `/out`-prefixed.

mrshu (commented):
Looks good overall, a couple small nitpicks and questions here and there.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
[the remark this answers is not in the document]
Adman: Yeah, @mrshu is right, the directory should be provided in advance.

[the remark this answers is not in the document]
dbscoach: Okay, so kill path generation and get output path from os.environ['video_path'] entirely?

run-docker.sh
dbscoach: We had already named the test files bot1/bot1.py etc. I'm not married to the concept, but this way I can already run two halves correctly. I'll adapt to whatever you specify for submission. I changed the name of the PR from "Implementation" to "Proof of Concept", none of this needs to stay.

run-webots.sh
dbscoach: Not ask any blocking questions, if it were to ask you someting it will just always go with the default. I don't think it does anything here, but better safe than sorry if we change something and it starts to.

Dockerfile
dbscoach: Yes, 1234 is the default port for x3d-live-streaming, that's a remnant from my first docker-compose-file. It wouldn't even actually get exposed if we didn't have network=none because the run-command doesn't map it anywhere. I'll remove it.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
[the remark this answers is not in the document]
Adman: Not really, check line where `output_path` function is called and create `output_prefix` like this

```python
if automatic_mode:
    output_prefix = output_path(Path('/out'), TEAM_BLUE, TEAM_YELLOW)
else:
    output_prefix = output_path(Path('reflog'), TEAM_BLUE, TEAM_YELLOW)
```

run-docker.sh
mrshu: Thanks @dbscoach, it's all good, please do not read this in any other way than "I am extremely happy this even works and would want to thinker more with it :slightly_frowning_face:

run-webots.sh
mrshu: Makes sense -- thanks @dbscoach !

Dockerfile
mrshu: Thanks @dbscoach I was literally just curious :slightly_frowning_face:

dbscoach:
This has been superseded by the docker-tournament-run-branch

PR #69, opened 2021-02-01
Proof of concept for running tournament using make

dbscoach:
Not pretty either, but it does run games and output video files with teams by their IDs in them.


controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -12,7 +12,9 @@
 automatic_mode = True if "RCJ_SIM_AUTO_MODE" in environ.keys() else False
 
 TEAM_YELLOW = environ.get("TEAM_YELLOW_NAME", "The Yellows")
+TEAM_YELLOW_ID = environ.get("TEAM_YELLOW_ID")
 TEAM_BLUE = environ.get("TEAM_BLUE_NAME", "The Blues")
+TEAM_BLUE_ID = environ.get("TEAM_BLUE_ID")
```
Adman, line 17: We should assign these IDs some default values, otherwise the `output_prefix` path would contain `None`.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -12,7 +12,9 @@
 automatic_mode = True if "RCJ_SIM_AUTO_MODE" in environ.keys() else False
 
 TEAM_YELLOW = environ.get("TEAM_YELLOW_NAME", "The Yellows")
+TEAM_YELLOW_ID = environ.get("TEAM_YELLOW_ID")
```
mrshu, line 15:
```suggestion
TEAM_YELLOW_ID = environ.get("TEAM_YELLOW_ID", 999)
```

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
mrshu, line 17:
```suggestion
TEAM_BLUE_ID = environ.get("TEAM_BLUE_ID", 990)
```

mrshu (commented):
Looks good @dbscoach -- thanks for putting this together.

I think we could (and probably should) extract all these scripts out of this repository. They are only used for running the tournament and not really relevant for all those that would like to try out the simulator on their own.

dbscoach:
This has moved to the rcj-soccer-sim-tournament-utils repo.

Issue #74, opened 2021-02-08
Pass the score of the first halftime to the next one


Adman:
Via environment variable

Issue #17, opened 2020-12-07
Auto gameplay pipeline 


aquahika:
Pack in docker?
Input several student's .py files and output game playback video and statistics.

aquahika:
Running a webots on Docker is introduced at the official website:
https://www.cyberbotics.com/doc/guide/installation-procedure#installing-the-docker-image

On my environment with nvidia GPU, after installing nvidia-docker2 package it works well: 
![image](https://user-images.githubusercontent.com/2618822/102357070-4efffe80-3ff1-11eb-90c0-4ff4127df897.png)

dbscoach:
Is there any easy way to run this headless without losing GPU acceleration? I tried and failed...

mrshu:
@dbscoach @aquahika We should probably get together and figure out whether we can actually use GPU acceleration. I am happy to provide a server for testing purposes as well, if that would be of any help.

Would something like that work for you?

Thanks!

PR #78, opened 2021-02-21
Changes to add ability to control recording, half-time starting scores, pass in team names etc. for tournament run

dbscoach:
Also contains stuff @mrshu merged into the docker-tournament-run branch from (I think) master.


dbscoach:
Fixes #74 and in combination with the rcj-soccer-sim-tournament-run repo also fixes #17

Adman:
Before reviewing, I'll squash the commits and rebase on top of master so the diff shows the changes correctly.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -7,11 +7,14 @@
...
 
-automatic_mode = False
+from os import environ
 
-TEAM_YELLOW = "The Yellows"
-TEAM_BLUE = "The Blues"
+automatic_mode = True if "RCJ_SIM_AUTO_MODE" in environ.keys() else False
 
+TEAM_YELLOW = environ.get("TEAM_YELLOW_NAME", "The Yellows")
+TEAM_YELLOW_ID = environ.get("TEAM_YELLOW_ID")
```
Adman, line 15: This should be `environ.get("TEAM_YELLOW_ID", 0)`, because otherwise we end up having `None` in the name of json file.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -7,11 +7,14 @@
...
+TEAM_BLUE = environ.get("TEAM_BLUE_NAME", "The Blues")
+TEAM_BLUE_ID = environ.get("TEAM_BLUE_ID")
```
Adman, line 17: This should be environ.get("TEAM_BLUE_ID", 1), because otherwise we end up having None in the name of json file.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -28,10 +31,14 @@ def output_path(
         directory.mkdir(parents=True, exist_ok=True)
 
     p = directory / Path(f'{team_blue}_vs_{team_yellow}-{now_str}')
+    if automatic_mode:
+      match_id = environ.get("MATCH_ID", "")
```
Adman, line 35: All environment variables should be put together on top of file, so we know what we are using.
Also, use 4 spaces, not 2, please.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -107,7 +108,10 @@ def __init__(
         self._update_positions()
 
         self.ball_reset_timer = 0
-        self.score_blue = self.score_yellow = 0
+        score_blue = os.environ.get("TEAM_B_INITIAL_SCORE", "0")
+        self.score_blue = int(score_blue) if score_blue != "" else 0
+        score_yellow = os.environ.get("TEAM_Y_INITIAL_SCORE", "0")
+        self.score_yellow = int(score_yellow) if score_yellow != "" else 0
```
Adman, line 114: All of these env variables should be read in rcj_soccer_referee_supervisor.py and we pass their values to the referee during initialization.

Adman (changes requested):
Thanks for the PR! I fixed commits so there are not changes, which have already been merged to master. However, I have some comments regarding the PR.

Why do we need to copy the demo code to three folders? It makes in messy and hard to make any changes.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
Adman, line 17: We should create some kind of validator that in case of automatic mode, we expect the ID, otherwise raise an exception.

Adman: And we should set some default value because we end up having `None` in the name of json file

Adman:
@dbscoach FYI, I did some small changes to the python files.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
mrshu: @Adman or we could perhaps fail completely if this this is not provided?

dbscoach:
> @dbscoach FYI, I did some small changes to the python files.

Thank you very much. I guess I'll go break more things then ;)

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
dbscoach: I just put broken code into the supervisor on purpose. It can crash without webots exiting which would clog the tournament simulation pipeline. How complex would it be to exit out of webots if we don't get valid variables?

Adman: @dbscoach ah, you are right, webots will be running even if supervisor script fails.

dbscoach:
@mrshu @Adman I have updated to now record optionally record mp4 or x3d formats, controlled via environment variable REC_FORMAT and this apparently immediately shows up in this PR, but I didn't want to create a branch of a branch.

controllers/rcj_soccer_referee_supervisor/recorder/recorder.py
```diff
@@ -44,19 +44,22 @@ def get_resolution(self):
...
         # API details for movieStartRecording
         # https://www.cyberbotics.com/doc/reference/supervisor?tab-language=python#wb_supervisor_movie_start_recording
-        self.supervisor.movieStartRecording(filename,
-                                            width,
-                                            height,
-                                            quality=100,
-                                            codec=0,
-                                            acceleration=self.fastforward_rate,
-                                            caption=False)
+        if 'mp4' in formats:
```
Adman, line 53: Let's create a class `class RecordingFormats(Enum)` in conts.py for this, and add `mp4` and `x3d` there.

controllers/rcj_soccer_referee_supervisor/rcj_soccer_referee_supervisor.py
```diff
@@ -7,33 +8,49 @@
...
+output_prefix = output_path(
+    directory,
+    TEAM_BLUE_ID,
+    TEAM_YELLOW_ID,
+    MATCH_ID,
+    HALF_ID,
+)
 reflog_path = output_prefix.with_suffix('.jsonl')
-video_path = output_prefix.with_suffix('.mp4')
+video_path = output_prefix.with_suffix('') # recorder sets suffix depending on filetype
```
Adman, line 53: I do not like this, because `with_suffix('')` does nothing. I would instead create another method, similar to `start_recording`, but this time call it like `start_animation_recording`.

Adman (changes requested):

Adman:
@mrshu I made some changes to create video assistants based on the format provided. Could you please check my changes?

@dbscoach If you find some time, could you please test if it's working?

Thanks!

dbscoach:
@Adman It is working, only thing it was complaining about is suffixes having to start with a period (so I changed them to .html and .mp4 from html and mp4).

controllers/rcj_soccer_referee_supervisor/recorder/recorder.py
```diff
@@ -26,11 +27,10 @@ def __init__(self,
             raise TypeError("Unexpected supervisor instance")
 
     def create_title(self):
-
         if self.output_path == "":
             # When output path is not specified
             time_str = datetime.datetime.now().strftime("%Y-%m-%d-%H:%M:%S")
-            return "{}/{}.mp4".format(str(Path.home()), time_str)
+            return "{}/{}.{}".format(str(Path.home()), time_str, self.output_suffix)
```
Adman, line 33: there will be two dots in the suffix. I'll fix it

Adman (approved):

Adman:
@aquahika Since we have been touching Video recorder assistant, would you have a time to check it please?

controllers/rcj_soccer_referee_supervisor/referee/consts.py
```diff
@@ -1,12 +1,12 @@
...
-
+import os
 
 class Team(Enum):
     BLUE = "B"
     YELLOW = "Y"
 
 
-MATCH_TIME = 10 * 60  # 10 minutes
+MATCH_TIME = os.environ.get("MATCH_TIME", 10 * 60)  # 10 minutes unless otherwise specified
```
Adman, line 9: This is not good. `consts` are ment to be constants -- no changes from outside.
So, let's call this `DEFAULT_MATCH_TIME = 10 * 60`, extract `MATCH_TIME` in main file and if the env variable does not contain anything, then fall back to `DEFAULT_MATCH_TIME` and pass it to `Referee` instance via `match_time` argument.

Adman (changes requested):

controllers/rcj_soccer_referee_supervisor/referee/consts.py
Adman: Fixed.

Adman (approved):

controllers/rcj_soccer_referee_supervisor/referee/consts.py
dbscoach: Oops. My bad, didn't realize this would immediately materialize on the PR. I was just testing and pushed just to move the code to the other machine. I'll fix it before the next push.

dbscoach:
Now that we have published should we merge this branch into master so that teams trying to run our simulator get code that will run by default?

Adman:
@dbscoach Yeah, let's merge it.

PR #81, opened 2021-03-03
Add notification whether a goal is scored

### LINKED ISSUE #73: Notify robots that goal was scored
Via the message sent by supervisor

Adman:
* closes #73

### DIFF
--- a/controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
+++ b/controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
@@ -286,14 +286,14 @@ class RCJSoccerSupervisor(Supervisor):
 
         Returns:
             str: the packed packet.
-
         """
 
         assert len(robot_translation) == len(robot_rotation)
 
         # X, Z and rotation for each robot
         # plus X and Z for ball
-        struct_fmt = 'ddd' * len(robot_translation) + 'dd'
+        # plus True/False telling whether the goal was scored
+        struct_fmt = 'ddd' * len(robot_translation) + 'dd' + '?'
 
         data = []
         for robot in ROBOT_NAMES:
@@ -307,6 +307,10 @@ class RCJSoccerSupervisor(Supervisor):
 
         data.append(ball_translation[0])
         data.append(ball_translation[2])
+
+        # Add Notification if the goal is scored and we are waiting for kickoff
+        # The value is True or False
+        data.append(self.ball_reset_timer > 0)
 
         return struct.pack(struct_fmt, *data)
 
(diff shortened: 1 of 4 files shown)


dbscoach (approved):
Looks good to me, have you run it yet?

Adman:
@dbscoach yes I did ;)

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
```diff
@@ -308,6 +308,9 @@ def _pack_packet(
         data.append(ball_translation[0])
         data.append(ball_translation[2])
 
+        # Add Notification if the goal is scored (True or False)
+        data.append(self.ball_reset_timer > 0)
```
mrshu, line 312: @Adman This is relatively nasty, this essentially mean the notification tells you whether we are in the time between when the goal gets scored and the new kickoff gets staged. I think calling it `is_goal` may be a bit dangerous in this regard (it is not set to true only in the "tick" when the goal gets scored).

Can you think of a name that's a bit more appropriate? I guess we should also explicitly describe this in the docs.

Thanks!

docs/docs/how_to.md
```diff
@@ -67,6 +67,7 @@ class MyRobot:
             "x": unpacked[3 * N_ROBOTS],
             "y": unpacked[3 * N_ROBOTS + 1]
         }
+        data["is_goal"] = unpacked[3 * N_ROBOTS + 2]
```
mrshu, line 70: @Adman is there a serious reason why leaving `3 * N_ROBOTS` would be better than abstracting it out in here? Asking mostly out of curiosity :)

Thanks!

mrshu (approved):
Just a small comment, nothing major though :)

Thanks for putting this together!

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Yeah, we should describe entire "packet" string in docs.

would something like `waiting_for_kickoff` or something similar work?

docs/docs/how_to.md
Adman: Yeah, I'll change the docs too make it more clear.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
mrshu: @Adman that does indeed sound very nice -- thanks!

docs/docs/how_to.md
Adman: Done.

controllers/rcj_soccer_referee_supervisor/referee/supervisor.py
Adman: Renamed.

PR #82, opened 2021-03-12
call lack of progress if object leaves the field

### LINKED ISSUE #77: Lack-of-Progress the ball if it leaves the field or increase tick rate so it doesn't anymore
Ball can pass through the wall if it does so in one tick and eventually score a goal (due to moving sideways enough to be behind the goal line)

Adman:
* fixes #77


mrshu (approved):
Thanks @Adman !

From the looks of it, we should really start putting some tests together, otherwise this will be wholly unmanageable in the future.

Adman:
@mrshu Yeah. Created an issue for that.

PR #86, opened 2021-03-16
world: Desynchronize the Robots

### LINKED ISSUE #50: Deal with robots that enter infinite loop
When a robot is not responding, end the match and the opponent wins 10:0.

mrshu:
* Previously, infinite loops resulted in halting the whole Webots
  application.

* As suggested in
  https://github.com/cyberbotics/webots/issues/2859#issuecomment-800042125
  setting synchronize synchronization to `FALSE` on all the robots ought
  to alleviate this problem.

* Fixes #50.

### DIFF
--- a/scripts/templates/soccer.wbt.template
+++ b/scripts/templates/soccer.wbt.template
@@ -515,6 +515,7 @@ DEF B1 Robot {
     mass 0.2
   }
   controller "$controller_b1"
+  synchronization FALSE
 }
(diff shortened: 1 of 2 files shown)


Adman (changes requested):
@mrshu Please fix also the template file in scripts folder.

mrshu:
Thanks for the pointer @Adman, should be done now.

Adman (approved):

PR #88, opened 2021-03-20
fix: fixed retyping robot id to int

Adman:


mrshu (approved):
I was contemplating creating a custom wrapper around `controllers.Robot` but I guess if this is all we need, it would be rather premature.

PR #89, opened 2021-03-20
fix: make supervisor synchronous

Adman:
* this should fix the time to speed up when
  the simulator runs at maximum speed
* from docs: "for the robot competitions
  generally the Supervisor controller is
  synchronous while the contestants
  controllers are asynchronous"


mrshu:
@Adman Looks good to me -- thanks!

Can you please link the docs to this commit? Or even better, can we link to them in the "world file"?

mrshu (approved):

Adman:
@mrshu Linked in the commit to keep the world file clean. (Adjusted also template file)

PR #91, opened 2021-03-24
fix: stop ball movement after relocation

who576:
* fixed ball moving after relocation


Adman:
@who576 Thanks for the PR. I'm discussing the bug with Webots here https://github.com/cyberbotics/webots/issues/2899
If we do not fix it, I'll merge this workaround then.

Adman (approved):

PR #94, opened 2021-04-21
fix checking if the recording is ongoing

Adman:


mrshu:
@Adman Let's at least mention https://docs.python.org/3/tutorial/classes.html#private-variables in the commit message or something like that -- this is one heck of a deep dive that had to happen in here!

mrshu (approved):

PR #103, opened 2021-05-07
add check-submission script

Adman:

### DIFF
--- a/docs/docs/code_submission.md
+++ b/docs/docs/code_submission.md
@@ -100,6 +100,10 @@ Note that there is no `controllers/` folder anymore -- the `.ZIP` file only
 contains a single `robot/` folder and a single text file called `team_name.txt`
 (and optionally the `logo.png` image).
 
+**If you would like to check whether your `.ZIP` file conforms the required
+structure, we recommend checking it by running `check-submission.py` script,
+which can be found in `scripts` directory of the `rcj-soccer-sim`.**
+
 ## Uploading your submission
 
 Once you have your `.ZIP` file ready, the only thing left is to submit it to
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -22,3 +22,14 @@ $ python generate-soccer-world.py \
 
 Among other things, this allows us to switch team sides without changing the
 internal logic of the simulation.
+
+## check-submission.py
+
+A script used for checking whether the zip file, which is to be submitted,
+conforms the required structure. If you want to check your submission,
+run the following command:
+
+```bash
+$ python check-submission.py <path_to_submission.zip>
+```
+
--- a/scripts/check-submission.py
+++ b/scripts/check-submission.py
@@ -1,0 +1,41 @@
+import os
+import sys
+import zipfile
+
+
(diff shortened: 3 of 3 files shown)


mrshu:
@Adman I have just one question: can we reimplement this in JavaScript? 

I guess it should be possible -- things like `zip.js` seem to support it out of the box: https://gildas-lormeau.github.io/zip.js/demos/demo-read-file.html -- especially since we don't look into the files and only check their existence.

That way, we could essentially embed this check inside the docs (at least on modern browsers) and make it easy even for people for whom running Python from the command line may be difficult (a lot of Windows users I imagine).

This script is still worth having, as it will allow us to do automatic checks on the submissions but I believe a web-client-only version may be even more interesting in here :slightly_smiling_face:

docs/docs/code_submission.md
```diff
@@ -100,6 +100,10 @@ Note that there is no `controllers/` folder anymore -- the `.ZIP` file only
 contains a single `robot/` folder and a single text file called `team_name.txt`
 (and optionally the `logo.png` image).
 
+**If you would like to check whether your `.ZIP` file conforms the required
+structure, we recommend checking it by running `check-submission.py` script,
+which can be found in `scripts` directory of the `rcj-soccer-sim`.**
```
mrshu, line 105: @Adman we could maybe include a link afterwards

scripts/README.md
```diff
@@ -22,3 +22,14 @@ $ python generate-soccer-world.py \
 
 Among other things, this allows us to switch team sides without changing the
 internal logic of the simulation.
+
+## check-submission.py
+
+A script used for checking whether the zip file, which is to be submitted,
+conforms the required structure. If you want to check your submission,
```
mrshu, line 29:
```suggestion
conforms to the required structure. If you want to check your submission,
```

scripts/check-submission.py
```diff
@@ -0,0 +1,41 @@
...
+import zipfile
+
+
+def fail(message):
+    print("FAIL: {}".format(message))
+    sys.exit(1)
+
+
+if len(sys.argv) != 2:
+    fail("usage: python3 {} <path to zip>".format(sys.argv[0]))
```
mrshu, line 12: @Adman does this need to be `python3`? I guess it may be confusing to some.

scripts/check-submission.py
```diff
@@ -0,0 +1,41 @@
...
+    fail("robot/ folder not found in {}".format(submission))
+
+if "robot/robot.py" not in namelist:
+    fail("robot.py not found within robot/ folder in {}".format(submission))
+
+if "team_name.txt" not in namelist:
+    fail("team_name.txt not found in {}".format(submission))
+
+for entry in namelist:
+    if entry != "robot/" and entry.endswith("/") and entry.count("/") == 1:
```
mrshu, line 34: @Adman very nicely done -- took me a while to figure out this is checking the number of top-level directories :)

mrshu (approved):

Adman:
Moved the script to utils instead.

PR #106, opened 2021-05-17
submission_check: Fix paths

mrshu:
* Set absolute paths in `submission_check.md` so that it actually works
  in the end.


mrshu:
@Adman couldn't find a better solution quickly -- I guess that `mkdocs` thing could allow us to do some URL rewriting but for now this should do.

PR #108, opened 2021-05-24
docs: Change main program name to robot.py

felipenmartins:
Since the submission check expects a main program called robot.py, maybe it would be wise to use the same name in the explanation. Could you see if my proposed changes make sense?


mrshu (approved):
Thanks @felipenmartins !

Adman (approved):

PR #115, opened 2021-10-12
Decrease depth to prevent robot falling over

Adman:
* regenerate world with newest webots version


Adman:
@mrshu This should help but I'm afraid it is once again some webots problem with physics.

Adman:
Yeah, it should not cause any hassle. Merging.

PR #125, opened 2022-01-06
Update Webots to R2022a

### LINKED ISSUE #124: Update Webots to R2022a
This requires converting some of the objects as well as fixing the x,y,z in the code.

Adman:
* convert the world data to keep the old behaviour
* fix compass to return angle 0 when the robot is
  facing opponent's goal
* closes #124

### DIFF
--- a/scripts/templates/soccer.wbt.template
+++ b/scripts/templates/soccer.wbt.template
@@ -1,15 +1,14 @@
-#VRML_SIM R2021b utf8
+#VRML_SIM R2022a utf8
 WorldInfo {
   info [
     "A simple RCJ Soccer game simulated using Webots."
   ]
   title "RCJ Soccer Sim"
   optimalThreadCount 8
-  coordinateSystem "NUE"
 }
 Viewpoint {
-  orientation 0.7514920402341333 -0.6392670342257186 -0.16308700873151558 5.42412
-  position 1.25037 1.41585 1.7555
+  orientation -0.3718171524084895 -0.11726675761807528 0.920869433053669 3.706903255785043
+  position 1.7483931840718845 1.2701664687979481 1.965968373123659
 }
 TexturedBackground {
 }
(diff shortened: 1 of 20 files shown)


Adman:
I kept the old NUE coordinate system not to confuse participants before the competition.

Do you think we could still convert everything to "ENU" and FLU axis orientation?

mrshu:
> I kept the old NUE coordinate system not to confuse participants before the competition.

@Adman I believe pretty much everyone will start coding their robots from scratch. If that ends up being the case, I would go with whatever makes the most sense (and what ends up being the default in Webots) rather than backwards compatibility.

Do you see some other compelling counter-argument I might be missing?

scripts/templates/soccer.wbt.template
```diff
@@ -1,4 +1,4 @@
-#VRML_SIM R2021b utf8
+https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022a
```
mrshu, line 1: @Adman I know literally nothing about this but shouldn't this line say `#VRML_SIM R2022a utf8`? In all honesty, I am not convinced Webots would be able to parse this. Perhaps we could add one more line with this link somewhere into the file (say the second line)?

mrshu (changes requested):

Adman:
Ok, then I'll try to convert it to the new coordinate system and fix the code.

Adman:
@mrshu The conversion should be done. I fixed the referee as well so now it is working with `[x,y]` instead of `[x,z]`.

scripts/templates/soccer.wbt.template
```diff
@@ -1942,5 +2006,4 @@ Robot {
   ]
   controller "rcj_soccer_referee_supervisor"
   supervisor TRUE
-  synchronization TRUE
```
mrshu, line 1945: @Adman does this mean `synchronization` is set to `TRUE` by default? Looking at your commit (https://github.com/RoboCupJuniorTC/rcj-soccer-sim/commit/2dbabb77aaa1c4c7452c650d9d51eab387ce5bbb), it does sound like something we'd like to keep set to `TRUE` for the foreseeable future, right?

mrshu (approved):
Thanks for going through all this @Adman, I very much appreciate you taking the time to look through all the places where the `z` axis used to be used.

I guess there isn't an automated way of ensuring something didn't slip through, right? We'll just have to play a ton of games and test in prod :slightly_smiling_face: 

Other than that, I have just a single question regarding `synchronization` -- other than that, this looks good to me.

scripts/templates/soccer.wbt.template
Adman: @mrshu Yeah synchronization is set to `TRUE` by default, that's why webots deletes the property everytime the world is saved. I reverted it back so that it is explicitly set to `TRUE`. Thanks for pointing out.

Adman:
@mrshu I quickly tested the lack of progress, penalty area check, goals etc so we have at least some confidence it works. Let's merge this and then test it out during the demo competition.

PR #129, opened 2022-03-30
Set TIME STEP to 32 for robot controllers, ball and supervisor

Adman:
* this should sync everything with world's
  `basicTimeStep = 32`


controllers/rcj_soccer_team_yellow/robot1.py
```diff
@@ -41,8 +41,8 @@ def run(self):
                 # If the robot has the ball right in front of it, go forward,
                 # rotate otherwise
                 if direction == 0:
-                    left_speed = 5
-                    right_speed = 5
+                    left_speed = 7
```
mrshu, line 44: @Adman any particular reason for these `7`s?

mrshu (approved):

controllers/rcj_soccer_team_yellow/robot1.py
Adman: @mrshu `5` appeared to be quite slow after I changed `TIME_STEP` to 32. 7 is better.

Issue #130, opened 2022-04-17
Robot relocation bug


RichoM:
Hi, I've been testing the simulator a lot lately and I've noticed a problem with the relocation of robots. A lot of times when the supervisor relocates a robot, it ends up flipped upside down (or to the side, in the worst case).

To show what I'm talking about I recorded a short video. These are all instances of robots being flipped (and a few close calls) during just one match: 

https://youtu.be/4vt4nLYVCsI

- Webots version: r2022a
- OS: macOs Mojave version 10.14.6
- Simulator: Latest (2278b6d48b612231d27884aa013389be6c5a84c9)

I'll see if I can test in a windows computer and I'll let you know if the problem occurs there as well.

Adman:
Hello @RichoM, thanks for reporting the problem.

The code responsible for moving the object is located [here](https://github.com/RoboCupJuniorTC/rcj-soccer-sim/blob/master/controllers/rcj_soccer_referee_supervisor/referee/supervisor.py#L77-L87)

```python
tr_field = self.robot_translation_fields[robot_name]
tr_field.setSFVec3f(position)
self.robot_nodes[robot_name].resetPhysics()
```

It might be possible that the robot is running at high speed while being relocated. Since I'm not able to reproduce the problem so often, could you please change the order of these two lines in supervisor's code and test it? Thanks!
```python
tr_field.setSFVec3f(position)
self.robot_nodes[robot_name].resetPhysics()
```

RichoM:
Thanks for the quick response. I tested your proposed fix but it didn’t seem to have any effect.

I think you’re right in that it could be related to the robot speeds so I tried to change the team code to detect lack of progress and stop the motors a few moments before being relocated. That seemed to help a little but not as much as I hoped.

To help reproduce the problem more easily I made the following team controller:

```python
from controller import Robot

TIME_STEP = 32

robot = Robot()

left_motor = robot.getDevice("left wheel motor")
right_motor = robot.getDevice("right wheel motor")

left_motor.setPosition(float('+inf'))
right_motor.setPosition(float('+inf'))

while robot.step(TIME_STEP) != -1:
    left_motor.setVelocity(10.0)
    right_motor.setVelocity(10.0)
```

With the code above running for both teams I can reproduce the bug in almost 100% of relocations (at least on my computer).

RichoM:
[...]

PR #133, opened 2022-04-28
Workaround for the relocation bug

RichoM:
Workaround for issue #130


Adman:
@RichoM Thanks for the fix. Could you please fix the failing build, squash your commits into one force-push to your branch?

RichoM:
I think I fixed it. Let me know if not

Adman (approved):
Thanks for the fix! Let's hope it will be fixed in Webots soon.

PR #140, opened 2022-12-31
Reset - End of the Year Challenge 2022 Final Submission

resetteam2023:
### Team Reset: End of The Year Challenge 2022 Submission

**Main Changes:**

- All robots run on 4 wheels instead of 2 wheels
- A dribbler is added to the robot
- A kicker is added to the robot
- The radius of the ball is reduced from 0.021 to 0.015
- The basicTimeStep of the world is reduced from 32 to 8
- A new shape is created for the robot
- The coulombFriction and the forceDependentSlip values between the wheels of the robots, the ball, and the field are changed
- The density of the robot is set to 17000 (the mass of the robot, which was previously 0.2, is no longer used)
- The center of mass of the robot is lowered by 0.01 along the z-axis

_The current world only works for the 2022b version of Webots and not the 2023a version (which was released not long ago). A world that is compatible with the 2023a version of Webots has been made. However, the world is yet to be finalized due to some issues._


worlds/soccer/Robot Shape.obj
```diff
@@ -0,0 +1,193 @@
+# WaveFront *.obj file (generated by Autodesk ATF)
+
+mtllib Robot Shape.mtl
```
Adman, line 3: @resetteam2023 My first question regarding the shape of the robot is: Would it be possible to change the cube shape into cylindrical one?

resetteam2023:
Hello @Adman, 

It would be possible to change the shape of the robot to a cylindrical shape rather than a cube shape. However, we have not tested it out yet as we believe that it may futher complicate the physics of the world.

If you would like to change the shape of the robot to be cylindrical, let us know and we will update the world for you.

Thank you!

Team Reset

Adman:
@resetteam2023 If it is not too much complicated to test cylindrical shape, let's try it out.

I noticed one issue when I run webots R2022b. It prints some DRIBBLER related as well as some physics related warnings

`WARNING: DEF B1 Robot > DEF DRIBBLER HingeJoint > Solid: Webots has detected that this solid is light and oblong according to its inertia matrix. This belongs in the physics edge cases, and can imply weird physical results. Increasing the weight of the object or reducing its eccentricity are recommended.` 

`WARNING: The current physics step could not be computed correctly. Your world may be too complex. If this problem persists, try simplifying your bounding object(s), reducing the number of joints, or reducing WorldInfo.basicTimeStep.`

Do you have an idea if it is possible to fix that?

Thank you!

worlds/.soccer1.wbproj
```diff
@@ -0,0 +1,9 @@
+Webots Project File version R2023a
```
Adman, line 1: Do we need these `.soccerX.wbproj` files? What are they used for?

Adman:
Can you please remove `.DS_store` file?

controllers/rcj_soccer_ball/rcj_soccer_ball.py
```diff
@@ -8,5 +8,5 @@
 data = [True]  # Packet cannot be empty
 packet = struct.pack("?", *data)
 
-while robot.step(32) != -1:
+while robot.step(8) != -1:
```
Adman, line 11: Why is the time step reduced to 8? Did 32 caused any problems?

mrshu (commented):
Very nicely done @resetteam2023 -- kudos for putting this together!

I've left a couple of comments here and there, feel free to respond at your convenience :slightly_smiling_face:

worlds/.soccer1.wbproj
resetteam2023: Hello @Adman,

Sorry, we were not aware that there were those files in the folder. We assume that it may have been a file that was created by the Webots program when we were trying to making our Webots world with the R2023a version. Our guess is that it would not be a problem to simply delete those files.

On a side note, we originally created the world to work with the R2023a version rather than the R2022b version of Webots. However, due to issues with obtaining the correct values for the ball direction (the issue was only present in the Mac version of Webots but not the Windows version, for some reason), we ended up changing our code to work with the R2022b version. As a result, there may likely be unneccessary files created by the R2023a version of Webot's program.

Team Reset

controllers/rcj_soccer_ball/rcj_soccer_ball.py
resetteam2023: Hello @Adman,

Yes, since the physics computation is now a lot more complicated than before, changing the value for the basic time step of the world would drastically affect the way that robots move and perform. After testing, we found that setting the basic time step of the world to 8 is the most ideal. When we set the basic time step to 32, the physics-related warning message `WARNING: The current physics step could not be computed correctly. Your world may be too complex. If this problem persists, try simplifying your bounding object(s), reducing the number of joints, or reducing WorldInfo.basicTimeStep.` appeared more frequently. On the other hand, when we set the basic time step to a value lower than 8, the robots no longer moved as we expected.

Let us know if you have any other questions regarding this issue!

Team Reset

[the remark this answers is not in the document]
resetteam2023: Hello @mrshu,

Sure! We have responded to the question in a comment above (in a conversation with @Adman). However, if you would like us to discuss about the question in more detail, feel free to let us know!

Team Reset

controllers/rcj_soccer_team_blue/rcj_soccer_robot.py
```diff
@@ -191,5 +204,14 @@ def get_sonar_values(self) -> dict:
             "back": self.sonar_back.getValue(),
         }
 
+
+    def kick_ball(self, state, velocity=20):
+        if state == 1:
+            self.kicker_motor.setVelocity(velocity)
+            self.kicker_motor.setPosition(0.042)
```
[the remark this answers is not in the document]
resetteam2023: Hello @mrshu,

Yes, the value of 0.042 here represents the furthest position that the kicker could move to without moving beyond the frame of the robot (the greater the value, the further the center of the kicker is away from the center of the robot). 

When we were making the kicker in the Webots world, we only made sure that the kicker position had to be between 0 and 0.042. While this means that the kicker could be moved to positions between 0 and 0.042, the kicker would not be kicking rthe ball as hard in those cases. 

If you would like, we could add a parameter to the `kick_ball` function that would allow the kicker to be moved to specific positions between 0 and 0.042. 

Thank you!

Team Reset

controllers/rcj_soccer_team_blue/robot1.py
```diff
@@ -38,18 +41,37 @@ def run(self):
...
+                # Dribble the ball at a velocity of 50
+                self.dribbler_motor.setVelocity(50)
+
+                # Kick the ball for 0.5 seconds every 5 seconds
+                if 5000 < kick_time < 5500:
+                    self.kick_ball(1, 20)
+                elif kick_time > 5500:
+                    kick_time = 0
+                else:
+                    self.kick_ball(0, 20)
```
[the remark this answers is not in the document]
resetteam2023: Hello @mrshu,

If `kick_ball` is never executed with `state = 0`, after `kick_ball` is executed with `state = 1`, the kicker would remain in the "outward" position (the center of the kicker would be at a position of 0.042). This means that the robot would not be able to kick the ball again and would not be able to dribble the ball.

Let us know if you have any further questions!

Team Reset

resetteam2023:
Hello @Adman and @mrshu,

We just committed and pushed new changes to this forked repository. In this comment, we would like to address these new changes as well as some of the questions above that we have yet to answered (that relate directly to these new changes). 

1. We changed the shape of the robot from a cube shape to a cylindrical shape. In order to keep the width and depth of the ball-capturing zone the same, we made the radius of the circular cross-section of the robot's new shape equal to 0.0425. While we changed the radius of the robot, we kept the height of the robot the same (equal to 0.075) to make the mass of the robot as similar as possible to before (the mass of the robot is dependent on the fixed value that we set for the robot's density and the robot's volume). Additionally, since we realized that a robot dribbling the ball would capture more than half of the ball in its ball-capturing zone when the robot's shape is simply a cylinder (with radius of 0.0425), to not have to change the radius of the robot (would affect the width and depth of the ball-capturing zone), the front side of the robot directly above the ball-capturing zone is kept flat.

2. We removed all the unnecessary files including the `.soccerX.wbproj` files and the `.DS_store` files. As we expected, removing those files did not affect the simulation in any significant way.

3. While we assumed that the physics computation for the Webots world would become even more complicated if we made the shape of the robots cylindrical rather than cubic, we figured through testing that the physics computation for the Webotrs world actually became simpler when we made the shape of the robots cylindrical (which is odd because a cylindrical shape is theoretically much more complicated than a cubic shape). After making the shape of the robots cylindrical, we have not seen the warning message about the physics computation being too complex (`WARNING: The current physics step could not be computed correctly. Your world may be too complex. If this problem persists, try simplifying your bounding object(s), reducing the number of joints, or reducing WorldInfo.basicTimeStep.`). However, if that warning message continues to show up (even though we have not seen them after running the simulation multiple times), please let us know. We have tried to deal with this issue for a while already and are still trying to figure out the exact reason behind why that warning message shows up.

4. In regards to the warning message of the dribbler being too light (`WARNING: DEF B1 Robot > DEF DRIBBLER HingeJoint > Solid: Webots has detected that this solid is light and oblong according to its inertia matrix. This belongs in the physics edge cases, and can imply weird physical results. Increasing the weight of the object or reducing its eccentricity are recommended.`), we think that the best way to get rid of the warning message would be to simply increase the mass of the dribbler. However, although there is a warning for the dribbler being too light, since we do not think that the dribbler being too light has caused problems to the simulation (the physics computation being too complex issue, on the other hand, does cause problems) and changing the mass of the dribbler would require us to change many other parts of the simulation, we have decided to simply ignore the warning message. If you would still like us to get rid of that warning message, though, feel free to let us know.

If you have any other questions or proposals for changes, let us know and we will try to respond to them as soon as possible. Thank you!

Team Reset

worlds/.soccer1.wbproj
Adman: Thank you.

Adman:
@resetteam2023 Thanks for your updates. I have checked it and it looks very nice!

I've got further questions:
1. On each robot, only front sonar (ultrasound) sensor is visible. Can you move also rear, left and right ones outside of the robot, so they are visible as well?
2. I think the simulation is a bit slower (even the seconds decrease slower). When I run the simulation with old world, I'm getting the simulation speed around 0.98x and running it in fastest mode increases to ~2.30x. Experimenting with the time step a bit I found out that `TIME_STEP = 32` makes the simulation run again at ~0.98x (however, the fastest mode is just ~1.40x. Do you think we could revert the time step back to 32?
3. How much of an effort would it be to fix the DRIBBLER warning by increasing its mass?

Adman:
@resetteam2023 I've been playing with the robot a little and found out that the wheel is doing "8" while rotating. Do you think that this could be possible to fix?

https://user-images.githubusercontent.com/865063/211917568-084d1433-f924-4e1d-b188-446697fa23f7.mp4

resetteam2023:
Hello @Adman ,

Thank you for pointing out the issue and sorry for getting back to you so late. We were on Chinese New Year break so did not get the opportunity to look into those issues yet.

However, we will look over the issues you mentioned (sonar positioning, slow simulation speed, light dribbler mass, and strange wheel movement) in the next few days and try to respond as soon as possible. We would like to mention, though, that we have tried to increase the speed of the simulation but were not able to increase it as much as we would have liked. We will continue to see if there are any other ways to increase the speed of the simulation, but as the simulation is now a lot more complicated than before, it may be difficult to do so.

Thank you and please let us know if there are any other parts of the simulation you would like to fix. 

Team Reset

Adman:
Hello @resetteam2023. How are things going? Let me know if I can help somehow. Looking forward to seeing the updates.

resetteam2023:
Hello @Adman ,

Everything is going fine! We have increased the mass of the dribbler (and changed a lot of other parameters of the world to accomodate for it) and moved the ultrasonic sensors out so that they are visible. We have realized that there is an issue with the wheel movement and are working on it. We will commit the code and update you as soon as we get it to work (hopefully in a day or two). 

Thank you!

Team Reset

resetteam2023:
Hello @Adman ,

We just committed new code onto the repository. The changes that we made in the code should fix the strange wheel movement, remove the warning about the dribbler mass being too little, and make the ultrasonic sensors visible. Hopefully, all the issues (other than the simulation running too slow) are solved, but if there are any other issues, please let us know. We apologize that it took so long for us to make these changes.

Thank you!

Team Reset

Adman:
@resetteam2023 No worries, thank you for the update! Ultrasound sensors look OK except the front one, where there is a gap between the robot and the sensor (see the screenshot below). Dribbler warning is gone and wheels rotate normally -- that is great :+1: . I'll keep playing with the robots and test different movements.

It would be really nice if we could make the simulation faster, because the one second in simulation takes almost 2 seconds in reality. We should make it similar otherwise simulating matches would take a lot of time.
I would also like to ask if by any chance we could update the world to the newest version R2023a. How do you see that?

image

Adman:
@resetteam2023 Hello, I was playing with the dribbler a little bit and found out that robot is moving backwards when the velocity of motors if set to 0 and robot is dribbling the ball. Is it the weight of the robot being too low that could be causing this behavior? I guess physical robots most often stay at one place while dribbling the ball. Do you think we could fix this?

https://user-images.githubusercontent.com/865063/220465187-6eebf463-d423-4a17-948e-65567b873ebd.mp4

resetteam2023:
Hello @Adman ,

Thank you for pointing out the issues. 

We think that the fix to those issues would be to increase the mass of the robot. However, increasing the mass of the robot can cause a lot of problems for the simulation (including the robot slowing down, moving strangely, or at worst, becoming unable to move). In order to accommodate for those problems, we would have yo further complicate the physics of the simulation, which would slow down the simulation even more. However, we will try to work to fix those issues as soon as possible (we are currently traveling so we might take around a week or so) and let you know the results.

Thank you.

Team Reset

Adman:
@resetteam2023 Alright, if it slows the simulation down then I guess it should be fine. However, we should really fix the time step issue. Is that possible?

resetteam2023:
Hello @Adman ,

We have found a solution that will allow the simulation to run at a faster speed. The current solution (and the only one so far) that we have is to increase the TIME_STEP for the ball and the referee while making the TIME_STEP for the robots the same as the basicTimeStep. By doing so, a 10-minute simulation on our iMac could finish in about 3 minutes. While this solution does allow the simulation to speed up significantly, the issue that this solution presents is that the ball's values are no longer accurate (as the ball's values are updating slower), which goes back to the problem regarding TIME_STEP that was resolved prior to the simulation challenge last summer. A suggestion that we have will be to revert back to the rules that were used in 2021 where the ball's coordinates were given at all times rather than determined using the IR sensors. We think that not only would changing this rule potentially fix the issue, but it would also make gameplay more interesting. With the arc, the dribbler, and the ability for robots to move in all directions implemented this year, it is way too easy for teams to simply have two robots trap the ball and drive into the goal. 

On a side note, near the end of 2021 or early 2022, as you may already know, we started experimenting with implementing an arc to the robot. While we were not able to get the arc implemented in time then, we eventually ended up not only successfully implementing the arc, but also the dribbler, the kicker, and the ability for the robot to move in all directions (pretty much all the capabilities that the robot has now except for the ultrasonic sensors). In fact, what we created for this end of the year challenge was modelled off of what we started then. Interestingly, we went back to check the simulation that we built using the 2021 Webots version and found that not only was the simulation a lot smoother (we did not have to mess with the TIME_STEP and the robots ran well even when the basicTimeStep was set to 32), but that the simulation ran at around the same speed as the old simulation (the one used in the past few years where robots ran on two wheels). In a 10-minute simulation on our iMac, the simulation could finish in just around 1 to 2 minutes. Unfortunately, the entire system of Webots changed between 2021 and 2022 and as much as we tried, we were not able to successfully emulate what we built using the 2021 Webots version to the 2022 Webots version (even when we made all the parameters of the world the same). However, it did give us reason to suggest going back to using the 2021 rules where data about the ball position is given to the robots instead of having the robots determined the ball position based on the IR values. If you are interested in taking a look at what we built using the 2021 Webots version, feel free to let us know.

Additionally, as for the ultrasonic sensors, the reason why the front ultrasonic sensor seems to not be on the robot is because we wanted each ultrasonic sensor to be the same distance away from the center of the robot. Since the front of the robot's cylindrical volume is cut out, in order to make each ultrasonic sensor the same distance away from the center of the robot, we have to either make the front ultrasonic sensor not be part of the robot or make the other ultrasonic sensors not visible. If having the each ultrasonic sensor the same distance away from the center of the robot is not your concern, please let us know and we will change the simulation accordingly and commit the updated Webots world.

Since we feel like there is a lot to discuss and that it would be best to finalize the simulation as soon as possible, if you think it would also be helpful, we suggest having an online meeting where everything could discuss about what we could change and what we need to do in more detail.

Thank you.

Team Reset

Adman:
Dear @resetteam2023, would you mind rebasing your branch on top of our master? I've updated Webots to R2023b version. Let's see if that may help with the TIME_STEP issues.

mrshu:
@codex review

PR #141, opened 2023-09-29
Update webots to R2023b

Adman:
* change inter-robot communication message format
  to stringified json

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
 This is the official repository of the RoboCupJunior Soccer Simulator. The
 simulator is based on [Webots](https://github.com/cyberbotics/webots) and this
 repository provides both the "automatic referee" (which implements the [Soccer
-Simulated Rules](https://github.com/RoboCupJuniorTC/soccer-rules-simulation))
+Simulated Rules](https://github.com/robocup-junior/soccer-rules-simulation))
 as well as a sample simulated team of robots with some basic strategy.
 
 ![Soccer Sim](./docs/docs/images/soccer_sim.png)
 
-*Learn more in the [documentation](https://robocupjuniortc.github.io/rcj-soccersim/).*
+*Learn more in the [documentation](https://robocup-junior.github.io/rcj-soccersim/).*
 
 # How do I try this out?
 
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -3,24 +3,24 @@
 This is the official repository of the RoboCupJunior Soccer Simulator. The
 simulator is based on [Webots](https://github.com/cyberbotics/webots) and this
 repository provides both the "automatic referee" (which implements the [Soccer
-Simulated Rules](https://github.com/RoboCupJuniorTC/soccer-rules-simulation))
+Simulated Rules](https://github.com/robocup-junior/soccer-rules-simulation))
```
mrshu, line 6: @Adman Thank you for doing this!

controllers/rcj_soccer_team_blue/rcj_soccer_robot.py
```diff
@@ -138,9 +131,12 @@ def get_new_ball_data(self) -> dict:
...
                 }
         """
-        _ = self.ball_receiver.getData()
+        _ = self.ball_receiver.getString()
         data = {
-            "direction": self.ball_receiver.getEmitterDirection(),
+            # TODO: Remove [:3] once
+            #   https://github.com/cyberbotics/webots/pull/6394
+            #   is released and Webots updated
+            "direction": self.ball_receiver.getEmitterDirection()[:3],
```
mrshu, line 139: very nice find @Adman!

docs/docs/communication_between_robots.md
```diff
@@ -38,38 +38,30 @@ this channel are going to receive it. So, if `robot1` sends a message,
...
-
-We need to know the type of each value we want to pack into a packet. Let's see a
-quick example. The values we want to send are `v1 = 3.14`, `v2 = 5` and `v3 = True`.
-It is obvious that `v1` is floating point number, `v2` is integer and `v3`
-is boolean. Knowing the type of each variable, we need to define the
-structure of the message.
+Webots supports sending the packet in different formats (string, bytes, list of doubles...).
+Since our example controllers as well as supervisor use strings, let's describe how to send it.
+
+[JSON](https://en.wikipedia.org/wiki/JSON) is a format that stores human-readable text in key-value pairs.
```
mrshu, line 44: @Adman For a while I was wondering whether using something like [`bson`](https://bsonspec.org/) wouldn't be a better fit for the embedded usecase but seeing how the SoccerSim gets targeted mostly at lowering the barrier to entry, JSON does seem preferable indeed.

mrshu, line 44: > key-value pairs

@Adman strictly speaking, this is not the case, right?

docs/docs/communication_between_robots.md
```diff
@@ -38,38 +38,30 @@ this channel are going to receive it. So, if `robot1` sends a message,
...
+In short, it is very similar to the `dictionary` that python uses. 
+Python3 contains a built-in library called [json](https://docs.python.org/3/library/json.html). It
```
mrshu, line 46:
```suggestion
Starting from version 3.0, Python contains a built-in library called [json](https://docs.python.org/3/library/json.html). It
```

docs/docs/communication_between_robots.md
```diff
@@ -38,38 +38,30 @@ this channel are going to receive it. So, if `robot1` sends a message,
...
+offers encoding python's dictionaries into json string and later decoding it back to dictionary.
```
mrshu, line 47:
```suggestion
offers encoding Python's dictionaries into a JSON string and later decoding it back to a dictionary.
```

docs/docs/communication_between_robots.md
```diff
@@ -139,7 +113,11 @@ a `while` loop for example**
 
 ```python
 while self.team_receiver.getQueueLength() > 0:
-    packet = self.team_receiver.getData()
+    packet = self.team_receiver.getString()
     self.team_receiver.nextPacket()
     # Do something with the packet
 ```
+
+**WARNING: Webots does not guarantee the order of the messages. Your robot controllers should not rely on the order.
```
mrshu, line 121:
```suggestion
**WARNING: Webots does not guarantee the order of messages. Your robot controllers should not rely on the order.
```

mrshu (approved):
A few small comments @Adman, otherwise this does look good to me -- thanks a bunch for figuring it out!

docs/docs/communication_between_robots.md
Adman: This is the first time I hear about bson.
I would stick with the json for its simplicity and secondly that it's builtin python3, so no third party dependencies get involved that you have to maintain.

mrshu: Sure, makes sense, it just was the first thing that crossed my mind when reading about this :slightly_smiling_face:

docs/docs/communication_between_robots.md
```diff
@@ -38,38 +38,33 @@ this channel are going to receive it. So, if `robot1` sends a message,
...
-is boolean. Knowing the type of each variable, we need to define the
-structure of the message.
+Webots supports sending the packet in different formats (string, bytes, list of doubles...).
+Since our example controllers as well as supervisor use strings, let's describe how to send it.
+
+[JSON](https://en.wikipedia.org/wiki/JSON) (JavaScript Object Notation) is an
+open standard file format and data interchange format that uses human-readable
+text to store and transmit data objects consisting of attribute–value pairs and
+arrays (or other serializable values).
+In short, it is very similar to the `dictionary` that Python uses.
```
mrshu, line 48:
```suggestion
In short, it is very similar to the `dict` data structure that Python uses. 
```

Adman: Thank you, changed.

Adman:
@mrshu Thank you for the review. Merging.

